PR #3, opened 2015-11-01
Fix :meth: usage

brettcannon:
Done primarily to fix a syntax error, but also normalized usage of all `:meth:`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+docs/_build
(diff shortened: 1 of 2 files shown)


dabeaz:
I'm not a Sphinx expert so doc changes much appreciated!

brettcannon:
There is actually a bunch of other stuff you could do, but it's usefulness is mitigated by the fact that there is no directive for async methods yet (opened https://github.com/sphinx-doc/sphinx/issues/2105 for that)

dabeaz:
The whole await/async syntax is uncharted territory on syntax highlighting and related features generally. I think in terms of documentation, it makes sense to preface each operation with an explicit "await" as that is what you would have to do in code.   Also, it's less verbose than typing "this method is a coroutine" in each description.

brettcannon:
Totally agree, just unfortunate it breaks linking.

PR #7, opened 2015-11-02
updating markdown in README

pwittchen:
Hi,

I've updated markdown in README and changed its name from `README.rst` to `README.md`. Now, markdown is consistent with GitHub conventions and code snippets have syntax coloring. I think, it's easier to read now.

You can see how it looks here:
https://github.com/pwittchen/curio/tree/fix-markdown-in-readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ curio - concurrent I/O
 
 Curio is a modern library for performing reliable concurrent I/O using
 Python coroutines and the explicit async/await syntax introduced in
-Python 3.5.   Its programming model is based on cooperative
+Python 3.5. Its programming model is based on cooperative
 multitasking and common system programming abstractions such as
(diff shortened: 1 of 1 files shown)


nchammas:
This seems like a frivolous change. What's wrong with rst?

And as far as conventions are concerned, many projects use reStructuredText over Markdown, including [Python itself](https://github.com/python/cpython/tree/master/Doc), [Cryptography](https://github.com/pyca/cryptography), [Fabric](https://github.com/fabric/fabric), and [asyncio](https://github.com/python/asyncio).

dabeaz:
I would like to keep the README in Restructured Text because that's the same format as the other documentation and I'm likely to cut-and-paste text between the docs and the top-level README.

pwittchen:
Okay. If it's consistent with another standard, we can leave it as it is. I'm closing this PR.

PR #13, opened 2015-11-24
remove additional argument for new_taks function

### LINKED ISSUE #12: Additional asterisk in new_task function definition
currently the definition of `new_task` is like below. However, i don't know how does the second parameter, the asterisk, means.  

    async def new_task(coro, *, daemon=False):
    '''
    Create a new task.
    '''
        return await _new_task(coro, daemon)

In fact, if we call the `new_task` function with an `additional_useless_argument`, just like below:

    # hello.py
    import curio

    async def countdown(n):
        while n > 0:
            print('T-minus', n)
            await curio.sleep(1)
            n -= 1

    if __name__ == '__main__':
        kernel = curio.Kernel()
        additional_useless_argument = 0
        kernel.run(countdown(10), additional_useless_argument)

error message is shown:  
    
    TypeError: new_task() takes 1 positional argument but 2 were given

andyxning:
Fix #12

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 docs/_build
+.idea/
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -630,7 +630,7 @@ async def sleep(seconds):
     '''
     await _sleep(seconds)
 
-async def new_task(coro, *, daemon=False):
+async def new_task(coro, daemon=False):
     '''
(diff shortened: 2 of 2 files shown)


dabeaz:
The daemon flag is a keyword-only argument to new_task() by design.  Specify as a keyword, not as a positional argument.

Issue #21, opened 2015-12-05
Use intersphinx for cross-linking to Python docs


brettcannon:
http://sphinx-doc.org/latest/ext/intersphinx.html

E.g., the `selectors` module should be automatically linked to the Python docs.

zoidbergwill:
I've started working on adding this [here](https://github.com/dabeaz/curio/compare/master...zoidbergwill:intersphinx?expand=1), I've done all mentions of the `selectors` module. 

I was considering adding links to the standard lib documentation for some marked down Python keywords, like `yield from` and `async`/`await` but I'm not sure if that's overkill.

brettcannon:
Any progress, @zoidbergwill ? Could you possibly send a PR for what you have for the selectors module so that others can help fill in other gaps?

zoidbergwill:
I wasn't sure what other keywords would be good to link, so I didn't add any others. I was waiting for suggestions, but should've asked more directly.

I opened a PR for the selectors stuff: https://github.com/dabeaz/curio/issues/26

PR #26, opened 2015-12-21
Add intersphinx support for the selectors module

zoidbergwill:
Closes https://github.com/dabeaz/curio/issues/21

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -29,7 +29,11 @@ import shlex
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = [
+    'sphinx.ext.intersphinx',
+]
+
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -29,7 +29,11 @@
 # Add any Sphinx extension module names here, as strings. They can be
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
-extensions = []
+extensions = [
+    'sphinx.ext.intersphinx',
+]
+
+intersphinx_mapping = {'python': ('https://docs.python.org/3.5', None)}
```
brettcannon: Should probably use `https://docs.python.org/3` as the link so that the links are always against the latest release instead of pegging against 3.5.

brettcannon:
LGTM except for the one URL change I suggested.

zoidbergwill:
Thanks @brettcannon for the review. I've fixed that one URL.

PR #29, opened 2015-12-24
Fix an error on Windows when no sockets are being select()'d

mafemergency:
This won't call select() if there are no sockets being checked, as on Windows this will raise a WindowsError(WSAEINVAL)

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -335,11 +335,12 @@ class Kernel(object):
         else:
             timeout = None
 
-        events = self._selector.select(timeout)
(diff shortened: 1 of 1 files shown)


dabeaz:
Need more information about this pull request.  To my knowledge, curio doesn't even poll sockets if there are no tasks listening.  So, I'm a bit confused about the circumstances under which this might arise.  I could be missing something though.

dabeaz:
Closing this because the code has shifted a fair bit since it was first reported. Should note that I am open to future pull-requests related to Windows.

PR #38, opened 2016-02-21
Fix Python executable in test, use sys.executable instead of hard coded string.

tds333:
Allows to start test with python3.5 if python3 default is set to another version.


dabeaz:
Good catch!

PR #44, opened 2016-03-26
Add node echo server.

cjrh:
I was curious about performance with node. You already had a neat setup for benchmarking so I added a node.js echo server.  On my oldish i7 MBP, the node echo server is doing ~19K requests per second, with `curioecho.py` doing around 13.5K requests per second. For comparison on my hardware, gevent is doing ~16K.

Running the node echo server is very easy: `$ node nodeecho.js`


dabeaz:
Awesome.  This will be something interesting to play with.

PR #73, opened 2016-07-23
Split all examples as Python scripts

GMLudo:
Hi,

To be easier to play with examples for newcomers, I've splitted all examples as Python scripts.
It permits to launch examples directly on the computer.

BTW, thanks a lot about the documentation, especially about tutorial and FAQ, it's well architectured.

As permitted by the 3-clause BSD license, I want to re-use the architecture for a new AsyncIO tutorial and FAQ. Obviously, I'll keep a reference to your work.

Have a nice week-end.

### DIFF
--- a/docs/examples/blocking_io.py
+++ b/docs/examples/blocking_io.py
@@ -1,0 +1,17 @@
+from curio import run, spawn, tcp_server
+
+async def echo_client(client, addr):
+    print('Connection from', addr)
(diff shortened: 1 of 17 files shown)


dabeaz:
Good idea on having tutorial examples split out.  I need to investigate the test failure further.  Might restructure it slightly.  Thanks!

PR #74, opened 2016-08-01
Make the examples a little more accessible

aenglander:
Please excuse the unsolicited pull request. I understand that this is a personal preference and I will totally understand if the pull request is denied.

Catch KeyboardInterrupt to not make the user think an error occurred when killing the server with CTRL+c
Make SSL cert and key file locations absolute paths to be able to run the example from a different directory

### DIFF
--- a/examples/conn.py
+++ b/examples/conn.py
@@ -19,4 +19,7 @@ async def main():
 
 
 if __name__ == '__main__':
-    curio.run(main())
+    try:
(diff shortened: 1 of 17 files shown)


dabeaz:
Seems reasonable.  Almost wonder if a message such as "shutting down" could be printed on Ctrl-C instead of pass.  Thanks!

PR #85, opened 2016-10-17
Task-local storage

njsmith:
Implement task-local storage for curio.

No tests yet, but everything else is there.

I added a nice example of the API in action to the tutorial -- probably that's the thing to look at first.

# Open questions

To highlight the two API decisions that are most likely to be controversial / I'm most uncertain about:

## Sync vs async

I made the interface totally usable from synchronous code. Aside from being weird, this does force certain decisions in the implementation -- in particular it requires some hoop-jumping to set up and maintain a global (thread-local) rendesvous point for locating the current task metadata. This might have performance implications -- I haven't run benchmarks. (Specifically, doing it this way means there's a bit of extra overhead at every task switch; on the other hand, I suspect but haven't checked that it makes accessing TLS values faster than it would be if we had to trap every time.)

The rationale for this decision is that a major, probably *the* major use case for TLS in curio is for storing contextual metadata for logging. And logging is a thing where you kind of need to make it work, even under adverse conditions. Like if you're trying to siphon out data from stdlib `logging` and into a better system, then you should do that and probably rewriting every third-party library to stop using stdlib `logging` is not a viable option -- but stdlib `logging` is sync-land, so if we made this an async API we'd be stuck. Also you probably don't want to ever find yourself in the situation where you want to stick some debug statement in some random leaf function somewhere to figure out why everything is on fire... but oops that leaf function is synchronous, so you can't write `await log.debug(...)` without first refactoring a whole call chain.

## Inheritance

I made it so that child tasks inherit the initial values of TLS data from their parents (as a simple one-time shallow snapshot). This is very unusual among TLS implementations, but for a system like curio where tasks are small, cheap, and common, then I think it makes a lot of sense. Other alternatives include: (1) not inheriting, (2) fancier inheritance, like having children keep a pointer to their parent and if a key isn't found in a child traverse the graph looking for it (so changes in parents propagate to children but not vice-versa).

This does add some overhead to task spawning, though I suspect that it's *very* small in the common case (though it scales with the number of TLS keys in use).

### DIFF
--- a/curio/__init__.py
+++ b/curio/__init__.py
@@ -11,6 +11,7 @@ from .queue import *
 from .workers import *
 from .network import *
 from .file import *
+from .tls import *
 
 __all__ = [ *errors.__all__,
             *task.__all__,
@@ -21,4 +22,5 @@ __all__ = [ *errors.__all__,
             *workers.__all__,
             *network.__all__,
             *file.__all__,
+            *tls.__all__,
            ]
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -19,6 +19,7 @@ from .errors import *
 from .errors import _CancelRetry
 from .task import Task
 from .traps import _read_wait, Traps
+from .tls import _enable_tls_for, _copy_tls
 
 # kqueue is the datatype used by the kernel for all of its queuing functionality.
 # Any time a task queue is needed, use this type instead of directly hard-coding the
@@ -415,6 +416,7 @@ class Kernel(object):
         # Add a new task to the kernel
         def _trap_spawn(_, coro, daemon):
             task = _new_task(coro, daemon)
+            _copy_tls(current, task)
             _reschedule_task(current, value=task)
 
         # Reschedule one or more tasks from a queue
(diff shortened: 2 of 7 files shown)


njsmith:
Added tests, but ATM they just hang due to #86, so I don't know if they're correct.

dabeaz:
Think I managed to introduce some conflicts with recent changes.  Inclined to merge though.

njsmith:
Rebased on master and fixed the tests. We'll see what travis says, but they pass locally now for me. (Except for `tests/test_file.py::test_readiter`, but that also fails for me on master -- I guess I should file a bug on that too huh :-).)

Anyway, I think it's ready for review/merge.

dabeaz:
I think the test_readiter test requires Python 3.5.2 or newer--there was some change in how asynchronous iteration was supported between 3.5.1 and 3.5.2.  (Suppose I ought to add a version check in the test).

njsmith:
Oh right, the thing about whether `__aiter__` is async or not. Yeah, that would explain it. It's not terribly hard to support both IIRC, but certainly annoying.

The travis failure is in the doc build:
```
/home/travis/build/dabeaz/curio/docs/reference.rst:204: WARNING: py:exc reference target not found: curio.TimeoutCancelledError
```
I don't think this is my fault :-)

dabeaz:
Wondering if you saw this:   https://docs.google.com/document/d/1tlQ0R6wQFGqCS5KeIw0ddoLbaSYx6aU7vyXOkv-wvlM/edit

njsmith:
I hadn't! That is super interesting, thanks!

The first thing I learned is that apparently my shallow-copy-on-spawn idea is the same one that C# went with some years ago, so I guess that's a vote of confidence.

curio/__init__.py
```diff
@@ -11,6 +11,7 @@
 from .workers import *
 from .network import *
 from .file import *
+from .tls import *
```
ZhukovAlexander: +1 for this. I was also a bit surprised by this. This should not be a issue for users, as they wan't interact with this module directly, but this will definitely confuse someone who is willing to explore the internals of curio.

njsmith:
Yeah, I was talking about this with @1st1 last night... Maybe we should
rename to "context-local storage"

PR #88, opened 2016-10-27
Implement reentrant lock

ZhukovAlexander:
This PR introduced the RLock implementation, similar to the `threading.RLock`.
This enables curio to allow users to re-acquire the lock if it is own by the 
same `curio.Task` instance. Here is an example use case:

``` python
import curio

async def coro(lck):
    async with lock:
        print('Critical section')

async def parent_coro(lck):
    async with lock:
        await coro()

async def main():
    await curio.spawn(parent_coro(curio.RLock()))

if __name__ == '__main__':
    curio.run(man())
```

If the above code was using a simple `curio.Lock`, it would end up in a deadlock, while with
a reentrant lock it works just fine.

### DIFF
--- a/curio/sync.py
+++ b/curio/sync.py
@@ -131,7 +131,7 @@ class RLock(_LockBase):
         self._count = 0
 
     async def acquire(self):
-        
+
         me = await current_task()
 
         if self._owner is not me:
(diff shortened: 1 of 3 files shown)


ZhukovAlexander:
Hi. This PR doesn't include the documentation update yet. I'm going to add this later if this PR will make it to the curio.

PR #97, opened 2016-11-10
Rename task-local storage -> context-local storage

njsmith:
To avoid confusion with Transport Layer Security and Thread-Local
Storage.

### DIFF
--- a/curio/__init__.py
+++ b/curio/__init__.py
@@ -11,7 +11,7 @@ from .queue import *
 from .workers import *
 from .network import *
 from .file import *
-from .tls import *
+from .cls import *
 
 __all__ = [ *errors.__all__,
             *task.__all__,
@@ -22,5 +22,5 @@ __all__ = [ *errors.__all__,
             *workers.__all__,
             *network.__all__,
             *file.__all__,
-            *tls.__all__,
+            *cls.__all__,
            ]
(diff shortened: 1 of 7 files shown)


defnull:
`cls` is often used to abbreviate "class". Why not just call the module `local`?

    > from curio.local import Local

njsmith:
I don't really care, feel free to bikeshed it out and let me know what you
decide :-). (Note that there's lots of references inside the code and docs
too though, not just the filename.)

imrn:
On 11/10/16, Nathaniel J. Smith <notifications@github.com> wrote:
> (Note that there's lots of references inside the code and docs
> too though, not just the filename.)

I'd recommend PyCharm as IDE.

This would be a non-issue with its Refactor->Rename operation.

It also provides a review (all files-locations) before
actually performing the rename.

dabeaz:
Hmmm.  Reviewing this.  All things equal, I think I would like to keep the terminology "task local storage" because of its similarity to "thread local storage" the similarity between Curio tasks and threads.

I agree that the TLS acronym should probably be avoided though.  I'm inclined to renamed to the file to `local.py`.   Certain functions such as `_enable_tls_for` might get renamed to `_enable_tasklocal_for`.

Give me your thoughts.  If no objections,  I can make the necessary changes.

njsmith:
@dabeaz: Makes sense to me.

dabeaz:
I've made this change.

PR #98, opened 2016-11-10
Make __aiter__ methods work on old and new Pythons

njsmith:


ZhukovAlexander:
Should we worry about backward compatibility at such an early stage? Especially for minor patch versions. I'm afraid, that we could end up making the same mistakes, as some other libs did when they were trying to be as backward compatible as possible.

dabeaz:
As a general rule, I don't like compatibility shims.  As an experimental project and the fact that the nature of async/await is evolving itself in core Python, I don't have any issue telling people that they need to use the most recent version of Python.   I consider it somewhat likely that I'll start using Python 3.6 only features once that's out of beta.

njsmith:
The main issue here is that requiring a specific *micro* version of Python
(3.5.2) is annoying compared to depending on a specific mini version (3.5).
A secondary issue is that this cleans up the code to handle __aiter__ in a
consistent, forward-looking way, so that when we do drop 3.5 support we
just have to delete the shim. (Notice that right now the code is
inconsistent about whether it hard-codes the 3.5.0 vs 3.5.2 behavior.)

On Nov 10, 2016 04:57, "David Beazley" <notifications@github.com> wrote:

As a general rule, I don't like compatibility shims. As an experimental
project and the fact that the nature of async/await is evolving itself in
core Python, I don't have any issue telling people that they need to use
the most recent version of Python. I consider it somewhat likely that I'll
start using Python 3.6 only features once that's out of beta.

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dabeaz/curio/pull/98#issuecomment-259683373>, or mute
the thread

njsmith:
To expand a bit on that, by "awkward" I mean: (1) people aren't used to micro version requirements, (2) tooling isn't set up for it (e.g. a wheel can require 3.5.0-or-better, but it can't require 3.5.2-or-better), (3) I happen to be lagging a bit on CPython updates so curio's test suite keeps failing for me locally, which is what annoyed me into submitting this patch :-).

skgbanga:
This was recently closed, but the example given in [Developing with curio](https://curio.readthedocs.io/en/latest/devel.html) currently produces a warning about `__aiter__` being awaitable. 

(also the countdown is from 5, instead of 4)

```
from curio import run

class AsyncCountdown(object):
    def __init__(self, start):
        self.start = start

    async def __aiter__(self):
        return AsyncCountdownIter(self.start)

class AsyncCountdownIter(object):
    def __init__(self, n):
        self.n = n

    async def __anext__(self):
        self.n -= 1
        if self.n <= 0:
            raise StopAsyncIteration
        return self.n

async def main():
    async for n in AsyncCountdown(5):
        print('T-minus', n)

>>> run(main)
T-minus 5    <---- this is wrong
T-minus 4
T-minus 3
T-minus 2
T-minus 1
>>>
```

Issue #96, opened 2016-11-09
Need a way to protect a region from cancellation (like "masking interrupts")


njsmith:
Here's an example of a simple bidirectional proxy using curio: https://gist.github.com/njsmith/4b4e75a9ab9e975705412f8f8e866948

The core idea is that we spawn two tasks to handle the copying in each direction:
```python
async def copy_all_bidir(sock1, sock2):
    await all_or_none([
        await curio.spawn(copy_all(sock1, sock2)),
        await curio.spawn(copy_all(sock2, sock1)),
])
```

The tricky part is managing these tasks. Basically with regards to exceptions/timeouts/cancellation/etc., we want it to act like the two `copy_all` tasks are both "part of" the `copy_all_bidir` task. So:
* If one of the tasks raises an exception, the other task should immediately be canceled, and the exception should propagate out of `copy_all_bidir`
* If both of the tasks complete, then `copy_all_bidir` should complete successfully
* If `copy_all_bidir` is cancelled, then both tasks should be canceled as part of `copy_all_bidir`'s cleanup

The `all_or_none` helper there tries to do this, but AFAICT it can't quite achieve the last requirement given curio's current APIs. The problem sequence is:

1. One of the tasks raises an exception
2. `all_or_none` catches this exception, and starts to cancel the other task
3. `all_or_none` itself gets canceled, thus canceling the cancellation and making everyone's head explode (and also leaking the other task)

In general this is a problem for cleanup code -- once you're executing the cancellation cleanup path you really don't want to get canceled again!

My suggestion for what it should look like is:
```python
async def all_or_none(tasks):
    try:
        async for task in curio.wait(tasks):
            await task.join()
    except:
        async with curio.defer_cancellation:
            for task in tasks:
                await task.cancel()
        raise
```
except that `curio.defer_cancellation` isn't a thing that exists right now.

So this issue is to propose that `curio.defer_cancellation` be a thing that exists.

(Guess how I'm distracting myself from election news!)

njsmith:
Oh also, this should probably *replace* the current `curio.timeout_after(None)` thing. I.e., `defer_cancellation` should protect against both timeouts and explicit calls to `Task.cancel()`, and then `timeout_after(None)` becomes unnecessary.

And the exact semantics I'm proposing are that if a cancellation arrives while `defer_cancellation` is in effect, then the cancellation exception gets raised from `defer_cancellation.__aexit__`.

[...]

PR #99, opened 2016-11-10
Add curio.defer_cancellation and Task.cancel_no_wait

njsmith:
curio.defer_cancellation is an async context manager which defers any
cancellations until after leaving the 'async with' block. See
gh-96.

It *should* also defer timeout exceptions, but it doesn't yet.

cancel_no_wait is like cancel, but it doesn't wait for the task to
exit. I needed it to write a test for defer_cancellation, and was just
wanting it anyway for some user code I was working on...

I'm not sure whether it makes sense to have both cancel() and
cancel_no_wait() -- maybe cancel_no_wait() should be renamed to just
become cancel()? But for now I left both in.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -444,16 +444,32 @@ class Kernel(object):
                     task.joining = kqueue()
                 _trap_wait_queue(_, task.joining, 'TASK_JOIN')
 
+        # Enter or exit a 'with curio.defer_cancellation' block:
+        def _trap_adjust_cancel_defer_depth(_, n):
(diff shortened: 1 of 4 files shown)


defnull:
I'd suggest adding a `wait=True` keyword only argument to `cancel()` instead of introducing a new method. The argument can later be modified to support a wait timeout.

dabeaz:
I'm somewhat inclined to add a `wait=True` keyword argument to `cancel()` instead of making it a separate method.  However, I'm going to merge this anyways.

Issue #101, opened 2016-11-10
[rfc] formalizing a split between "synchronous" traps and "blocking" traps


njsmith:
I'd like to modify the trap handling to formalize a distinction between:

* *synchronous traps*: traps which always return immediately, which never trigger a reschedule (so you are guaranteed that other tasks haven't touched any shared state), and are not cancellation points.

* *(potentially) blocking traps*: traps which might block, or might trigger a reschedule (`curio.sleep(0)`), and are cancellation points.

The two motivations are: (a) With my user trying to write cancellation-safe code hat on, it's very helpful if some kinds of calls are documented to be synchronous, and to know which ones those are and to have a uniform set of guarantees around them. This has even [bitten curio itself recently](https://github.com/dabeaz/curio/issues/86) -- `curio.Event` was assuming that `_queue_reschedule_function` was synchronous, but it wasn't; the fix was to make it synchronous. This strongly suggests that synchronicity is and should be a public API guarantee for some operations. (b) With my poking around in curio's internals hat on, I think this would let us simplify some of the code. In particular, synchronous traps could be implemented by calling a function to do the trap-specific work and then having some standard code to take that function's return value/exception, inject that into the invoking task, and reschedule it; this would be an improvement over the current situation where every trap implements that logic in an ad hoc way with `ready_appendleft` calls scattered around everywhere. It might also make things more efficient in that if we know that we're going to reinvoke the same task, we can skip enqueueing/looping/dequeueing entirely, and it potentially simplifies things like [alternative task scheduling policies](https://github.com/dabeaz/curio/issues/91) if the queues only have to handle real rescheduling, not trivial reinvoke-the-same-task "rescheduling".

Posting this here first because I want to get @dabeaz's opinion before I go refactoring a big chunk of curio's internals :-)

dabeaz:
Can you say a bit more about what you're thinking about?

One thing that's been on my mind generally is just the interaction between running tasks and the Curio kernel itself.  At the moment, I have taken a very strict approach by which all Task->Kernel interaction occurs via traps (and hence the await/async mechanism).  There is a certain consistency to doing it that way.   It makes it more like an operating system.  It also solves the problem of having to carry a reference to the kernel around (like needing the loop in asyncio).     Are you thinking that synchronous operations would occur in some other way?

njsmith:
Oh, sorry, no, I'm not suggesting changing the mechanics of the
task<->kernel API. I'm talking about standardizing the semantics of a
[...]

PR #103, opened 2016-11-12
Sync traps

njsmith:
Refactor trap internals to formalize the distinction between regular and "synchronous" traps. See #101 for motivation. Also convert the "obvious" traps to make them sync traps.

Specifically, the following traps are made sync:
* `adjust_cancel_defer_depth`
* `get_kernel`
* `get_current`
* `set_timeout`
* `unset_timeout`
* `queue_reschedule_function`
* `clock`
* `sigwatch`
* `sigunwatch`

The following traps in principle *could* be made sync, but either there's some complication or it isn't obvious whether those are actually the semantics we want, so I left them alone for now:
* `cancel_task`
* `spawn`
* `reschedule_tasks`

When switching traps to make them synchronous I also renamed them. This isn't at all mandatory, but since the name of the trap = the name of the trap handler function, and sync trap handlers have a different calling convention than regular trap handlers, I thought it was helpful to have that reminder in the name of which kind of handler we're implementing.

This also contains some small cleanups to regular trap handling and adds some missing docs.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -495,7 +495,7 @@ class Kernel(object):
                 _reschedule_task(current, value=time_monotonic())
 
         # Watch signals
-        def _trap_sigwatch(sigset):
+        def _sync_trap_sigwatch(sigset):
             # Initialize the signal handling part of the kernel if not done already
             # Note: This only works if running in the main thread
             if self._kernel_task_id is None:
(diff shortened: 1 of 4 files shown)


njsmith:
Just did the poor-form thing and added some small cleanups and an only-semi-related bug fix here... I can split them off if needed.

PR #110, opened 2016-11-14
Implement zero cost task switch

ZhukovAlexander:
This PR introduces the new switch utility. This should be more convenient to use, that `await sleep(0)`, makes API more semantically correct and also should be faster.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -530,6 +530,9 @@ class Kernel(object):
             current.state = 'TIME_SLEEP'
             current.cancel_func = lambda task=current: setattr(task, 'sleep', None)
 
+        def _trap_switch():
+            _reschedule_task(current)
+
         # Watch signals
         def _sync_trap_sigwatch(sigset):
             # Initialize the signal handling part of the kernel if not done already
--- a/curio/task.py
+++ b/curio/task.py
@@ -110,7 +110,7 @@ async def current_task():
 
 async def switch():
     """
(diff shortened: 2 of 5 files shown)


njsmith:
The implementation here doesn't work -- see the long comment in _trap_sleep
about why I removed the optimization where sleep(0) just called reschedule.

njsmith:
I'm not sure about the name, either... sleep(0) doesn't really bother me as
a spelling for this operation -- it's pretty idiomatic -- but having an
alias is fine too. If we do have an alias though I feel like the name
should be accurate, and "switch" is somewhat misleading, because this
doesn't actually guarantee that it switches tasks (or at least, it
shouldn't). Really what the operation is is just "let the scheduler know
that it *could* switch here, if it wants". So maybe "allow_switch" or
"schedule" or something would be a better name?

dabeaz:
Yes, the implementation here doesn't properly check for I/O as does `sleep(0)`.   If an alias to `sleep(0)` is desired, I think we could write it like this:

    async def schedule():
             await curio.sleep(0)

ZhukovAlexander:
Ah, ok, I missed this little detail. FWIW, [gevent uses the same approach](https://github.com/gevent/gevent/blob/master/src/gevent/hub.py#L166). Anyway, while digging into this stuff, I realized that we have a more [interesting issue](https://github.com/dabeaz/curio/issues/112).

PR #114, opened 2016-11-17
Pluggable scheduler queue proof-of-concept

ZhukovAlexander:
In this PR I propose a simple change to allow a pluggable task scheduler queue. I chose to add an argument to the kernel constructor instead of patching a global variable. Since other components of curio, like locks, semaphores etc, that utilize `kqueue`, do not need a particularly sophisticated internal queues to maintain waiters, this change will only impact a kernel scheduling. 

Feedback appreciated

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -34,12 +34,16 @@ kqueue = deque
 
 class Kernel(object):
     def __init__(self, *, selector=None, with_monitor=False, pdb=False, log_errors=True,
+                 scheduler_queue=None,
                  crash_handler=None):
         if selector is None:
             selector = DefaultSelector()
 
+        if scheduler_queue is None:
+            scheduler_queue = kqueue
+
         self._selector = selector
-        self._ready = kqueue()            # Tasks ready to run
+        self._ready = scheduler_queue()            # Tasks ready to run
         self._tasks = { }                 # Task table
         # Dict { signo: [ sigsets ] } of watched signals (initialized only if signals used)
         self._signal_sets = None
@@ -147,7 +151,7 @@ class Kernel(object):
         if self._thread_pool:
             self._thread_pool.shutdown()
             self._thread_pool = None
-        
+
         if self._process_pool:
             self._process_pool.shutdown()
             self._process_pool = None
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -190,9 +190,9 @@ class Kernel(object):
         selector_select = selector.select
         selector_getkey = selector.get_key
 
-        ready_popleft = ready.popleft
-        ready_append = ready.append
-        ready_appendleft = ready.appendleft
+        ready_pop = ready.popleft
+        ready_push = ready.append
+        ready_pushfront = ready.appendleft
         time_monotonic = time.monotonic
         _wake = self._wake     
 
@@ -231,7 +231,7 @@ class Kernel(object):
                     task.future = None
                     task.state = 'READY'
                     task.cancel_func = None
-                    ready_append(task)
+                    ready_push(task)
 
                 # Any non-null bytes received here are assumed to be received signals.
                 # See if there are any pending signal sets and unblock if needed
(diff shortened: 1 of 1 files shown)


njsmith:
Seems harmless, but I'm curious whether you've tried using this and found
it sufficient? I ask because I did some experimenting with weighted fair
queueing and I can't see how to make it work without a more substantial
restructuring. (In particular, the current system of alternating between
processing the ready queue and polling for I/O seems like it requires a
FIFO-like scheduler to work.)

ZhukovAlexander:
> whether you've tried using this and found
it sufficient?

Didn't quite get it. If you're talking about whether it'll be easy to supply different queues, like FIFO, LIFO, PriorityQueue etc., than this should be more than enough. You just pass in a queue, that provides the `deque` API but implement this API the way you need. Consider, for example, a simple priority queue:

```python
import heapq

class priority_queue:
    def __init__(self):
        self._items = []
    
    def append(self, item):
        heapq.heappush(self._items, item)

    def pop(self):
        return heapq.heappop(self._items)
```
You just then have to make comparable by some attribute, presumably their priority or some other statistic. We already have `Task._last_io` and `Task.cycles` to play with.

If you are talking about implementing more complex scheduling logic, this should also be possible, until you do all scheduling decisions only withing the queue itself.

Again, this is a POC, I assume we need something to start with. Also, I try to keep curio as simple as it is right now, that is why I tried to do it in the simplest way possible. Just duck type the deque and that's it.


Any firther suggestions are appreciated

njsmith:
Well, I'm not sure LIFO is an interesting scheduling algorithm in practice,
but as an example: to do LIFO properly, if a task is blocked on I/O and
then becomes runnable, it should be the next task run. But this isn't what
will happen if we just switch to a LIFO queue: instead we'll do this odd
hybrid of FIFO and LIFO scheduling, where we alternate between polling for
i/o and taking a batch of runnable tasks, and then running that batch in
LIFO order. The batching part is FIFO (all the tasks that were runnable
when we started the batch will get run before any tasks that were not
runnable at the start of the batch).
.
Priority based scheduling has similar issues: if a high priority task
unblocks, we should run it, not queue it up behind all the other tasks that
happened to be ready the polled for i/o. Also of course we would need to
start gathering and storing whatever information we want to use to compute
priorities.

ZhukovAlexander:
@njsmith , I get your point. I agree, for LIFO and FIFO we don't have much of a choice right now, as it's hard to tell which IO came first, and it's also hard to tell whether IO should come before sleeping tasks. At least. 

I somewhat disagree with your statement about priority queue though. If we use something like a binary heap, we can guarantee that the task with a higher priority will be scheduled before any other task with lower priority, even if we push it to the ready queue after other tasks. And note, that this property will be preserved even with the current event loop implementation, where we schedule IO and timers "in batches".

imrn:
I doubt doing fancy things on the task queues worth
the complexity and cost it brings. Even some embedded
systems operates with dumb cooperative scheduling of
small tasks and without priorities/preemption.
(Yes, I know you didn't propose preemption.)

This for a good reason: Not to carry the overhead and
complexity of fancy things at the very core.

I think compactness of Curio is the most valueable
feature we had to protect.

Infact I was the one who mentioned "Pluggable Scheduler"
for Hardened/Non-hardened Curio debate. But that was from
top of my head and it looks like more than 99% of the cases can
be covered with "Hardened App" approach which makes
"Hardened Curio" unnecessary.

Even it is doubtful "Pluggable Scheduler" will enable
Hardened Curio.. It is currently undecided. Probably
no decision is needed either.

njsmith:
True fair scheduling turns out to be super simple to implement -- you basically just need to keep a counter on each task; every time you run a task you increment the counter by how long it ran for, and when picking a task to run then run the one with the smallest counter value. There are a few details I'm skipping over and the theory of *why* this works turns out to be much more complicated than you'd expect (and I'm too lazy to type it all here :-)), but the actual implementation is extremely simple and pretty cheap.

It's not clear how much of a win it actually is in practice; probably its benefits are diluted both by the use of cooperative instead of pre-emptive scheduling, and by most curio programs being I/O bound rather than CPU bound. But in theory it seems like it could help servers provide better responsivity under load, e.g. like this question that glyph pointed to: https://stackoverflow.com/questions/40601119/python-ssh-server-twisted-conch-takes-up-high-cpu-usage-when-a-large-number-of/40627446

Issue #112, opened 2016-11-15
Kernel is vulnerable to starvation in some cases


ZhukovAlexander:
While working on [this PR](https://github.com/dabeaz/curio/pull/110) I was able to forge some code that made the event loop to completely skip the IO step. Here is the code that was used:

```python
import curio


lck = curio.Lock()

async def do_nothing():
    await curio.sleep(0)

async def worker(name):
    while True:
        async with lck:
            await curio.spawn(do_nothing())

async def main():
    await curio.spawn(worker('foo'))
    await curio.spawn(worker('bar'))


if __name__ == '__main__':
    curio.run(main())

```
The above program just runs forever without ever entering the IO polling phase. The root cause of it is that kernel uses greedy approach when running ready task, meaning, it will skip the IO completely, if there is at least one task ready to run ([relevant lines](https://github.com/dabeaz/curio/blob/master/curio/kernel.py#L623)). For reference, some libraries make sure to poll for IO for at least some minimal amount of time to give IO coroutines a chance to run.

dabeaz:
I'm somewhat less concerned with scheduling of locks/queues like this because I can't really envision a real-world situation where a program would just be spinning forever on a lock and never performing any I/O (a program like that would end up be 100% cpu-bound which is definitely not the kind of thing you'd want to run on Curio).     That said, I suppose the code could poll for I/O on rescheduling.   I don't know.  Open to thoughts.

njsmith:
I guess it might make sense to make lock acquisition a guaranteed schedule point, if we make a blanket rule that all potentially blocking operations are schedule points. But yeah, it's always going to be possible to write weird code that blocks the kernel -- the obvious example is just
```python
async def worker():
    while True:
         pass
```
:-)

So the question is more about coming up with more-nuanced guidance to users, so that they can look at a piece of code and make a clear prediction about whether it will be a problem or not, and where usually the answer is "not a problem".

njsmith:
Here's another starvation case. It's not realistic, but the effect is still a bit surprising. We have two tasks that talk to each other (here they happen to be identical, but you can imagine them implementing some sort of RPC pattern or something), and they actually *block* on each iteration -- i.e., are removed entirely from the ready queue. So you'd think they couldn't starve the kernel! Yet together they do, because they take turns putting each other back onto the ready queue, so that it never empties out:

```python
import curio

async def watchdog():
    i = 0
    while True:
        print("Event loop is live after {} seconds".format(i))
        i += 1
[...]

Issue #106, opened 2016-11-13
The eager read/write optimization leaves curio programs highly susceptible to event loop starvation


njsmith:
There was some discussion in #83 about removing the "eager write" optimization that curio does, where `await sock.send` will attempt to synchronously write to the socket directly, and only yield to the event loop if that fails. It turned out that the original reason we were considering removing it didn't actually apply, plus it significantly slows down the echo server microbenchmark, so for now the optimization survived.

But... I've just realized that actually the eager read/write optimizations create a serious correctness issue, in that they can cause ordinary-looking code to starve out the event loop in realistic situations.

Here's a [standard curio echo server](https://gist.github.com/njsmith/d6bb8d7fb517f6248b9fb940294d63bf), along with a particularly aggressive echo client. The test case is thta we run two copies of the client, and watch to see how well the server switches between servicing both of them. The server is instrumented to report how often it switches.

The server output pretty regularly contains lines like:
```
Switched after 220 loops (100.7 ms)
```
What this is saying is that one of the `echo_handler`s managed to complete 220 passes through its read/write loop before yielding to the kernel, so all other tasks were starved out for 100.7 ms.

I even saw a
```
Switched after 340 loops (171.6 ms)
```
without trying particularly hard. And in theory the only limit on how high those numbers can go is that at some point some scheduling quirk or something on my laptop causes one of the clients to be slow to read/write and makes the server block; under the right conditions it could go arbitrarily high.

@glyph has some [interesting comments on this optimization in Twisted](https://github.com/python/asyncio/issues/446#issuecomment-256498678), and how they removed it.

There's also an important difference between curio and twisted: in twisted, reads are *only* triggered by callbacks, so they always go through the event loop; it's only writes where you can even contemplate doing the eager-write optimization. So for them, this starvation problem is less likely to happen, because you can only hit it if you're doing an endless series of writes without any reads in between (and as glyph says, "/dev/urandom as a service has pretty limited utility"). In curio, we currently have *both* eager-read and eager-write optimizations, which makes the problem much easier to hit -- all kinds of realistic protocols are susceptible, starting with echo servers and proxies and moving on to things like an HTTP/1.1 server talking to a pipelining client.

[...]

Issue #91, opened 2016-11-04
[QUESTION] Experiment with alternative task scheduling policies


ZhukovAlexander:
IMHO, the `curio` library actually looks more like an operating system, rather than just a ***library for performing concurrent I/O***, which I think is intentional. Therefore many aspects of the OS development also apply to it. We already have a `Kernel`, a context of execution (aka **Task-local storage**), processes (aka ***Tasks***), system calls (aka `traps.py`)etc. 

So, one thing that I notice: the scheduling policy is a FIFO policy (this is actually the case for all major asynchronous libraries and frameworks), which is the simplest one, but by far not the optimal one. 

Given that `curio.Task` already has some execution stats (`Task.cycles`, `Task.state`, `Task. _last_io `), we can use this stats to come up with more sophisticated scheduling, like *Multilevel Feedback Queue*, which should improve the performance and usability under the real world workload. There were some experiments with threads scheduling already done: http://dabeaz.blogspot.com.au/2010_02_01_archive.html, which demonstrated a performance boost in a threaded environment, which should also apply to the coroutines world as well.

So the question is: should we try to use an advanced scheduling in `curio` and what benefits if any it'd bring?

dabeaz:
Quick answer:  Yes, we should make it possible to have alternative scheduling in Curio.

This is actually something that's been on the back of my mind.  However, in the interest of keeping the first few releases simple and manageable, it's been more of a back-burner project. 

Given the lack of true preemption, the flexibility of scheduling is probably going to be a bit more limited than in a real operating system.   However, I could definitely envision some scheme in which the kernel prioritizes tasks in some way.    It could also be interesting to explore priority scheduling in the context of soft-real time processing--it gets especially interesting if you mix it up with synchronization on locks, queues, and other things.   

If we go this route, I think my main concern would be making it modular in some way.   For example, the basic curio kernel could stick with FIFO as a default.  If you wanted alternative scheduling, it should be something that could be plugged in an optional feature.   I don't think it would be hard to do that.   Right now, the kernel uses collection.deque objects in a most minimal way.   That could certainly be expanded out into something more featureful.

ZhukovAlexander:
> If we go this route, I think my main concern would be making it modular in some way.

Sure, pluggable queue implementation looks like a way to go. This also aligns well with the [customizable kernel clock](https://github.com/dabeaz/curio/issues/92). 

[...]

PR #115, opened 2016-11-18
Unconditionally poll for I/O

ZhukovAlexander:
This PR forces even loop to poll for I/O unconditionally, even if there are tasks ready to run. This is done mitigate risks of starvation, so it makes the event loop more reliable (see #112), and also gives an opportunity to safely reschedule tasks. Performance of the event loop is not affected at all.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -619,30 +619,34 @@ class Kernel(object):
         # ------------------------------------------------------------
         while njobs > 0:
             # --------
-            # Poll for I/O as long as there is nothing to run
+            # Poll for I/O
             # --------
-            while not ready:
-                # Wait for an I/O event (or timeout)
+            if ready:
+                # set polling timeout to 0 if there is at least on task ready to run
+                timeout = 0
+            else:
                 timeout = (sleeping[0][0] - time_monotonic()) if sleeping else None
-                events = selector_select(timeout)
 
-                # Reschedule tasks with completed I/O
-                for key, mask in events:
-                    rtask, wtask = key.data
-                    if mask & EVENT_READ:
-                        rtask._last_io = (key.fileobj, EVENT_READ)
-                        rtask.state = 'READY'
-                        rtask.cancel_func = None
-                        ready_append(rtask)
-                        rtask = None
-                        mask &= ~EVENT_READ
-                    if mask & EVENT_WRITE:
-                        wtask._last_io = (key.fileobj, EVENT_WRITE)
-                        wtask.state = 'READY'
-                        wtask.cancel_func = None
-                        ready_append(wtask)
-                        wtask = None
-                        mask &= ~EVENT_WRITE
+            # Wait for an I/O event (or timeout)
+            events = selector_select(timeout)
+
+            # Reschedule tasks with completed I/O
+            for key, mask in events:
+                rtask, wtask = key.data
+                if mask & EVENT_READ:
+                    rtask._last_io = (key.fileobj, EVENT_READ)
+                    rtask.state = 'READY'
+                    rtask.cancel_func = None
+                    ready_append(rtask)
+                    rtask = None
+                    mask &= ~EVENT_READ
+                if mask & EVENT_WRITE:
+                    wtask._last_io = (key.fileobj, EVENT_WRITE)
+                    wtask.state = 'READY'
+                    wtask.cancel_func = None
+                    ready_append(wtask)
+                    wtask = None
+                    mask &= ~EVENT_WRITE
                     #if mask:
                     #    selector_modify(key.fileobj, mask, (rtask, wtask))
                     #else:
@@ -657,38 +661,38 @@ class Kernel(object):
                     ready_append(task)
                     '''
 
-                # Process sleeping tasks (if any)
-                if sleeping:
-                    current_time = time_monotonic()
-                    cancel_retry = []
-                    while sleeping and sleeping[0][0] <= current_time:
-                        tm, taskid, sleep_type = heapq.heappop(sleeping)
-                        # When a task wakes, verify that the timeout value matches that stored
-                        # on the task. If it differs, it means that the task completed its
-                        # operation, was cancelled, or is no longer concerned with this
-                        # sleep operation.  In that case, we do nothing
-                        if taskid in tasks:
-                            task = tasks[taskid]
-                            if sleep_type == 'sleep':
-                                if tm == task.sleep:
-                                    task.sleep = None
-                                    _reschedule_task(task, value=current_time)
-                            else:
-                                if tm == task.timeout:
-                                    if (_cancel_task(task, exc=TaskTimeout, val=current_time)):
-                                        task.timeout = None
-                                    else:
-                                        # Note: There is a possibility that a task will be
-                                        # marked for timeout-cancellation even though it is
-                                        # sitting on the ready queue.  In this case, the
-                                        # rescheduled task is given priority. However, the
-                                        # cancellation timeout will be put back on the sleep
-                                        # queue and reprocessed the next time around. 
-                                        cancel_retry.append((tm, taskid, sleep_type))
+            # Process sleeping tasks (if any)
+            if sleeping:
+                current_time = time_monotonic()
+                cancel_retry = []
+                while sleeping and sleeping[0][0] <= current_time:
+                    tm, taskid, sleep_type = heapq.heappop(sleeping)
+                    # When a task wakes, verify that the timeout value matches that stored
+                    # on the task. If it differs, it means that the task completed its
+                    # operation, was cancelled, or is no longer concerned with this
+                    # sleep operation.  In that case, we do nothing
+                    if taskid in tasks:
+                        task = tasks[taskid]
+                        if sleep_type == 'sleep':
+                            if tm == task.sleep:
+                                task.sleep = None
+                                _reschedule_task(task, value=current_time)
+                        else:
+                            if tm == task.timeout:
+                                if (_cancel_task(task, exc=TaskTimeout, val=current_time)):
+                                    task.timeout = None
+                                else:
+                                    # Note: There is a possibility that a task will be
+                                    # marked for timeout-cancellation even though it is
+                                    # sitting on the ready queue.  In this case, the
+                                    # rescheduled task is given priority. However, the
+                                    # cancellation timeout will be put back on the sleep
+                                    # queue and reprocessed the next time around.
+                                    cancel_retry.append((tm, taskid, sleep_type))
 
-                    # Put deferred cancellation requests back on sleep queue
-                    for item in cancel_retry:
-                        heapq.heappush(sleeping, item)
+                # Put deferred cancellation requests back on sleep queue
+                for item in cancel_retry:
+                    heapq.heappush(sleeping, item)
 
             # --------
             # Run ready tasks
--- a/examples/bench/curioecho.py
+++ b/examples/bench/curioecho.py
@@ -7,7 +7,7 @@ async def echo_handler(client, addr):
     print('Connection from', addr)
     client.setsockopt(IPPROTO_TCP, TCP_NODELAY, 1)
     while True:
-        data = await client.recv(1000000)
+        data = await client.recv(10000)
         if not data:
             break
         await client.sendall(data)


njsmith:
Huh, does this really not affect speed? It's certainly a neat and
straightforward solution if we can get away with it... I was just assuming
it would have to be more complicated :-). What benchmarks did you run?

ZhukovAlexander:
@njsmith , why overcomplicate things? Do you think polling for 0 seconds would actually affect performance?

dabeaz:
I'm not sure this would fix the starvation issue being discussed earlier.   I think the case that @njsmith was worried about involved I/O operations that completed immediately, never trapping into the kernel.   Since they never trap into the kernel, the scheduler simply wouldn't run.    This might explain the unchanged performance on the echo benchmark.

njsmith:
@dabeaz: I think it's intended to fix the oddities discussed in #112, not #106. Also it's a precondition to fancier scheduling policies, like discussed in #91 and #114. I guess it also makes the current scheduling algorithm into a stricter variant of FIFO, for whatever that's worth.

I think the only connection to #106 would be that this enables fancier scheduling policies, and if we had a fancier scheduling policy (e.g. one that explicitly reasons about fairness), and it was efficient enough, then we could potentially have a schedule point inside `Socket.recv`/`Socket.sendall`, but still kind-of preserve eager-read/write in the sense that we would only *actually* schedule away in read/write if scheduler determined that the task was hogging the CPU.

The I/O-polling thread I suggested in [this comment](https://github.com/dabeaz/curio/issues/106#issuecomment-260232149) was me trying to think of ways to get the semantics of this PR but while keeping the per-schedule overhead as low as possible.

njsmith:
@ZhukovAlexander: I was worrying about speed because I was thinking about how to convince @dabeaz to accept entering the scheduler more often, so I was trying to think about how to minimize the per-schedule overhead. This PR definitely does make `sleep(0)` (for example) somewhat more expensive than it was before. But maybe the difference is totally negligible.

dabeaz:
I'm somewhat less sure about this change and think it would require more study.

The primary use-case of Curio is in I/O bound applications.  The current scheduling is based on the goal of clearing the ready queue of all CPU processing as fast as possible (with the understanding that tasks aren't normally going to be doing a bunch of prolonged CPU-bound processing or doing pathological things with locks, queues etc.).   Under heavy load, the kernel would be processing things in bursts.  For example, I/O polling would release a group of tasks.  Those tasks would then run in quick succession, clearing themselves off of the ready queue.  While that's happening, more incoming I/O would be gathering in the background.  When you got around to the next I/O poll, you'd release another group of tasks to run and repeat. 

It's not entirely clear to me that polling for I/O on each reschedule would make this faster.  It might make it a lot slower.  Is it better to poll less often and get more ready tasks each time or is better to poll more often and only get a handful of ready tasks?   I just don't know.  I don't think the simple echo benchmark is enough to make a judgement one way or another about it though.

imrn:
Yeah. One should clearly finish ready tasks at the hand.

dabeaz:
I don't have enough of a gut feeling to know if this is a good idea or not.   However, in the interest of further experimentation, I'd consider a pull request that could turn it on as a optional feature.

imrn:
If starvation will occur due to some problem in the upstream tasks, polling less or more often, or polling in completely different order will NOT prevent it. It may just change
the time of the hit.  If a problematic combination is present somewhere, you'll get it eventually.

I think this will not solve the problem (because such problems are unrelated to here),
and bring unnecessary performance hit in real world.

imrn:
Please increase the # of unrelated simultaneously schedulable ops to
10s,100s,1000s.. and have the fun. The effect will be unnoticable for
low numbers. None for 1. Even this debate on performance is unnecessary.
Nothing can solve a starvation, lockup if it is there.
Infact, a tool for surfacing that condition as quickly as possible
will be very favorable. But it is not possible too.
There is no substitude for human touch.

njsmith:
>  Nothing can solve a starvation, lockup if it is there.

This PR does in fact solve the starvation/lockup cases described in #112. One could debate whether or not that's important to do, but it certainly does solve them.

imrn:
Changing the scheduling order did it by chance..
It is not guaranteed. In some other combination you'll
hit it again.

imrn:
Asymtotic equivalent of this proposition is this:
"Try all combinations of execution order"
Yes, this may be still be 'valid' probabilistic solution
in 'theory'..

njsmith:
I don't think you understand what this PR actually does?

imrn:
Doesn't such a polling scheme change the potential executing order?

njsmith:
Right now we only check for I/O if no tasks are on the `ready` queue. It's possible to get into situations where there is always at least one task on the `ready` queue, so any tasks that are waiting for I/O never run. This PR fixes that.

imrn:
So? You're providing somekind of priority to other potentially
runnable ops, hence "Changing The Order" and TRY to find a way
to continue..

What I'm saying is:

"Depending on the situation finding such a
combination is not guaranteed if any problematic
code is there. You may still lockup.
You are playing on your chances.."

I don't think you understand what my comment is about..

njsmith:
I guess the most productive way to move forward here would be for you to provide some code that spawns some tasks and:

* each individual task yields to the event loop infinitely often (no `while True: pass` or whatever)
* some task gets starved out and never runs

I believe that without this PR, it is possible to write such code (#112 has examples), and with this PR, it is no longer possible to write such code. AFAICT you disagree. If you can produce an example that would be very interesting!

ZhukovAlexander:
> in the interest of further experimentation, I'd consider a pull request that could turn it on as a optional feature

@dabeaz , great point about making it optional. I was actually planning to discuss some task execution throttling. According to the [Little's Law](https://en.wikipedia.org/wiki/Little's_law#Finding_Response_Time), the less items we have it the ready queue, the less the average response time will be. In case of an echo server this will minimize an average latency, which in some cases might be more important, than the overall throughput. So if we add a couple of line into the kernel, we can fine tune this:

```python
            while not len(ready) > self.MAX_TASKS_IN_READY:
                # Wait for an I/O event (or timeout)
                timeout = (sleeping[0][0] - time_monotonic()) if sleeping else None
                events = selector_select(timeout)

                # Reschedule tasks with completed I/O
                for key, mask in events:
```

Thus, we can kill two birds with one stone:

- gain control over the average kernel response time;
- [almost] mitigate risks of starvation etc.

ZhukovAlexander:
@njsmith , I know, this seems like a paradox, but this is true. The less tasks we schedule to run in a single turn, the more often a loop will cycle. Say, we already have 1000 tasks ready to run. If run them in one batch, the loop will be busy servicing this 1000 tasks, thus it kinda "blocks" until the ready queue is exhausted. That means, the loop is unresponsive to other tasks, like task cancellations and friends. This is also related to the ability to voluntarily preempt tasks (see  #106), since the faster the loop cycles, the faster it will continue to service the client after `sleep(0)` call. Basically, if we know, say from some benchmark, that the throughput of the loop is some N tasks per second, we shouldn't load the loop with more than that.

njsmith:
I still don't get it. The only way I can see that making the loop "cycle
faster" like this would help is if there's some latency-sensitive
operations that are performed once per cycle. What are those? Cancellation
for example just schedules a task to run; if the task being cancelled can't
get onto the ready queue then it doesn't matter how fast the ready queue
cycles. sleep(0) right now explicitly takes the task *off* the ready queue.

On Nov 20, 2016 12:32 PM, "Alexander Zhukov" <notifications@github.com>
wrote:

@njsmith <https://github.com/njsmith> , I know, this seems like a paradox,
but this is true. The less tasks we schedule to run in a single turn, the
more often a loop will cycle. Say, we already have 1000 tasks ready to run.
If run them in one batch, the loop will be busy servicing this 1000 tasks,
thus it kinda "blocks" until the ready queue is exhausted. That means, the
loop is unresponsive to other tasks, like task cancellations and friends.
This is also related to the ability to voluntarily preempt tasks (see #106
<https://github.com/dabeaz/curio/issues/106>), since the faster the loop
cycles, the faster it will continue to service the client after sleep(0)
call. Basically, if we know, say from some benchmark, that the throughput
of the loop is some N tasks per second, we shouldn't load the loop with
more than that.

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/dabeaz/curio/pull/115#issuecomment-261803483>, or mute
the thread

ZhukovAlexander:
Ok, imagine, we have 5000 tasks already scheduled to run. And we have 10000 connections to poll.
If we push them to the ready queue all at once, we end up with 15000 tasks. This might snowball, so we might end up with a pretty fat ready queue, like millions of tasks. It'll take a while for event loop to clean up all those tasks before it is available to do a next cycle. Sure, the loop will keep going, maybe even without any performance degradation, but we might eventually run out of memory or we will just have to wait for some time before the loop will react to other tasks.

dabeaz:
While I don't know about the existence of "millions" of tasks, the scenario described is not actually that unrealistic.  Image a heavily loaded "chat" server or something where there are 10000 active connections. Most of the connections are just sitting there not receiving much of anything, but when one of them actually does receive something, you now need to blast out a message like a status update to all of the connections at once.  Something like that could definitely drop a huge number of tasks on the ready queue all at once.

For something like this chat server, most of the pending work would involve making socket `send()` operations, almost all of which would likely complete immediately without blocking due to available buffers. Each task would then immediately go back to sleeping waiting for the next update.   In other words, even if you did have 10000 tasks on the ready queue all at once, they're not likely to be doing a whole lot of processing.  I could imagine the kernel cranking through the ready queue at a pretty fast clip. 

Here's my concern:  I'm not sure you want to be sticking a I/O poll operation in the middle of all of that fray.   For example, does processing everything on the ready queue now also require 10000 I/O polls?   I'm thinking it might be better to just get all of those tasks off the queue as fast as possible.

Although it's theoretically possible for tasks to starve the kernel doing weird things with queues and locks (i.e., ping-ponging data between tasks over a queue, etc.) all of this basically involves creating a situation where the kernel becomes CPU-bound.   Async is terrible at CPU-bound processing.  I just don't consider that to be the common case to be worried about (I don't think someone would want to intentionally create tasks that did tons of CPU processing without any actual I/O).

imrn:
One question: Do you consider starvation/lockup/what_ever_you_call/etc..
as a result of faulty app, or as a case that should be dealed, handled
and solved if possible?

Last question: What will be your response to this question if you
were running on threads?

dabeaz:
I don't think Curio has any responsibility to deal with starvation issues caused by pathological applications, no.  Specifically, the main use of Curio (and async more generally) is for I/O bound applications.  If you're writing code that isn't that, then you're probably going to experience a variety of performance issues.  I also don't think properly written applications should have their performance penalized on account of trying to accommodate every possible weird corner case.  (Full disclosure:  without measuring it, I have no idea what polling for I/O on every task schedule would do, but I have a hard time believing that it would be free for the last scenario described). 

As for threads, there are all sorts of bad things that can happen there too.   For example, CPU-bound work can't take advantage of multiple CPU cores (the GIL).   Improper use of thread synchronization primitives can cause deadlock.   Is it the responsibility of the thread library to handle all of these cases?  Not really.

Big picture:  I'm not sure there is a "one-sized fits all" policy for Curio to follow on a lot of these scheduling decisions.    As such, I'm much more interested interested in making sure that Curio provides developers with the basic tools necessary to fix such problems should it be demanded by the application at hand.  It's okay for it to have some knobs to turn.

imrn:
It shouldn't be that complicated David. At the very core "Async
Programming is just a Threads Replacement". No MORE, no LESS.
After realiably having this at the hand, I would not mind any value
add-ons or -'knobs'. But they shouldn't be distraction for the Large
and obvious target. Believe me: Nobody will care about the bells and
whistles once you hit the main target in the eye.
Curio has the potential. All it needs is a sharp focus.

dabeaz:
In your mind, is this specific scheduling concern part of a "sharp focus" or is it a knob to be turned?   Personally, I think it's a knob.

imrn:
It is a distraction. I'm surprised you asked this after all my comments
here and at other issues.

ZhukovAlexander:
Well, I think we should just abandon this until we have some real world cases that will require such treatment.

The last thing I want to get a resolution for is throttling. Do you guys think this is a valuable idea that deserves a dedicated PR? To me it looks nice and promising, but again, it's just a theory. We can at least start gathering some statistics about curio's task/second throughput. Thoughts?

dabeaz:
Distraction or not, I think it's important to emphasize that there are a lot of things going on all at once in Curio right now.   A big part of the project is about experimentation.   This whole async/await business is pretty new and Curio abuses it in all sorts of ways that probably weren't originally envisioned.   So, there's that.  Then there are concerns about making Curio more production hardened.  To that end, it's great that people like Nathaniel are thinking about evil corner cases, pathological code, and all sorts of related things.  I might not always agree with proposed solutions (or may want to defer), but I'm learning a lot from that and want to encourage it.   Last, but not least, the project is supposed to be fun.  I mean, who in their right mind goes out and creates a new async framework?  The `abide()` function?  What in the heck is THAT all about?    Given the newness and smallness of the code base, I don't mind people putting a few knobs on it and fooling around with stuff.  Best to do it now before there is all sorts of legacy code ;-).

imrn:
They are distraction even nuisanse in terms of mininal "Core" functionality.

"Clever" (even that will not work all the time) scheduling for those who even
doesn't know what really concurrency is about OR crippling the performance
upon request of the same audience is not a thing that you would do on
a tool called "Asyncronous Programming Library"..

Are there any clear plans to keep the "Core" strictly seperate from such
"Training Wheels"?

If there is, I'd even welcome and contribute to such nuisance..
And already I'm - if you haven't noticed until now..

dabeaz:
> Are there any clear plans to keep the "Core" strictly seperate from such
> "Training Wheels"?

I don't know if a "plan" has been formally stated, but my personal opinion is that the core Curio kernel should remain basically the same as it is now aside from bug fixes and cleanup--meaning that it's simple and fast.   If someone wants to play around with different scheduling policies, it should be done by creating custom code that lives outside the kernel.   Make a custom Socket object or something.

imrn:
Ok then let's put it in the main page in some form to set the
expectations of the audience with various degree of concurrency
experience and who is after perpetual motion machines.

njsmith:
@dabeaz:
> Full disclosure: without measuring it, I have no idea what polling for I/O on every task schedule would do, but I have a hard time believing that it would be free for the last scenario described

It's an interesting question, and probably worth measuring at some point just to know. For systems where we use a modern I/O API (epoll on Linux, kqueue on MacOS/*BSD, hopefully someday IOCP on Windows), then polling on every task schedule might turn out to be not free, but very cheap -- it should basically be as cheap as doing a cheap syscall. OTOH right now we wouldn't really get any benefit from doing this either.

@ZhukovAlexander:
> The last thing I want to get a resolution for is throttling. Do you guys think this is a valuable idea that deserves a dedicated PR?

I still don't understand how it could help, but if you put together a proof-of-concept demonstrating that it provides a nice win on some benchmark then I guess that'd be pretty convincing?

@dabeaz 
> the core Curio kernel should remain basically the same as it is now aside from bug fixes and cleanup--meaning that it's simple and fast. If someone wants to play around with different scheduling policies, it should be done by creating custom code that lives outside the kernel. Make a custom Socket object or something.

FWIW, there are two possible extensions to the current scheduling model that I've been thinking might eventually make sense to add to the core Curio kernel:

* "strict" priorities: i.e., tasks with low priorities don't run at all while there are higher-priority tasks runnable (like Linux SCHED_FIFO priorities). This is a somewhat common feature in other event loops ([example](https://developer.gnome.org/glib/stable/glib-The-Main-Event-Loop.html#G-PRIORITY-DEFAULT-IDLE:CAPS), [example](https://www.w3.org/TR/requestidlecallback/)), and useful for things like "I only want this background task to run if nothing else is going on", or "I want this interactive debugging tool to always take priority, even if the server is otherwise melting down trying to handle way too many incoming connections".

* weighted fair queueing: i.e., "optimally fair" scheuling in the sense that tasks that hold onto the event loop longer before yielding get scheduled less often, plus "soft" priorities that can be used to modulate this (like Unix "nice" values). The implementation for this is simple, fast, and elegant, and it might make curio-based programs less janky under load in [cases like this one](https://stackoverflow.com/questions/40601119/python-ssh-server-twisted-conch-takes-up-high-cpu-usage-when-a-large-number-of/40627446). Or it might not be useful at all -- I can't find any prior art for WFQ + cooperative scheduling. But worth experimenting with.

These are absolutely things where the next step is "play around with them outside of curio proper", and there are plenty of higher priority (heh :-)) things to worry about. But I mention them here because AFAICT they can't be implemented without modifying the kernel itself and otherwise integrating with core curio APIs. Which is fine, it just means that in the short term the method for playing around with them is to make a branch instead of writing a custom `Socket` object.

ZhukovAlexander:
Closing this in due to https://github.com/dabeaz/curio/pull/127

PR #116, opened 2016-11-19
[rfc] Finish splitting traps into blocking vs. synchronous

njsmith:
This implements the ideas described in gh-101, esp.

  https://github.com/dabeaz/curio/issues/101#issuecomment-261477568

Specifically, it makes it so all traps are either "blocking" or
"synchronous", where those have consistent, well-defined meaning:
blocking traps trigger scheduling, are cancellation points, and are
timeout points; sync traps do none of those things. The trap handler
calling conventions in the kernel are adjusted to follow this.

Branch also contains some logically independent cleanups that fit in
with the changes I was making anyway:

- Cleaned up cancellation logic so that it no longer requires the
  cancelling task to race with the victim task.

- defer_cancellation now defers timeouts as well as cancel calls.

- Renamed Task.cancel_no_wait to Task.cancel(blocking=False)

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -424,18 +424,13 @@ class Kernel(object):
             # This step indicates that we have managed any deferred I/O management
             # for the task.  Otherwise the run() method will perform an unregistration step.
             current._last_io = None
-            current.state = state
-            # current.cancel_func = lambda: selector_unregister(fileobj)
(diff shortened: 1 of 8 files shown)


dabeaz:
I'm going to merge this, but I think that _reschedule_task() should be blocking.  If priorities are introduced, it will be desirable to switch to the highest priority task at that point.  It might not be the currently executing task.

PR #124, opened 2016-11-21
Improve task locals API

ZhukovAlexander:
- use ```__slots__``` to decrease a memory footprint of ```curio.Local```
- ```vars()``` and ```dir()``` now work correctly when used with ```Local```
- forbid using ```Local``` outside of asynchronous context


njsmith:
Looks good to me, except that I'd leave in some of the test lines that
access __dict__ directly.

ZhukovAlexander:
@njsmith , no problem, let's keep those lines.

PR #129, opened 2016-11-24
Keep all possible tasks states in one place

ZhukovAlexander:
This is an attempt to make make it clear, which states are valid and possible for task to be in.
This will make it easier to reason about task state transitions within curio.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -17,7 +17,7 @@ log = logging.getLogger(__name__)
 
 from .errors import *
 from .errors import _CancelRetry
-from .task import Task
+from .task import *
 from .traps import _read_wait, BlockingTraps, SyncTraps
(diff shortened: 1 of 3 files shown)


dabeaz:
The state has always served a purely informational purpose in Curio and extends to many other things such as locks and queues.  For example, in `queue.py` there are states for `QUEUE_GET`, `QUEUE_JOIN` and so forth.    Just wondering if anyone has thoughts about whether or not all possible states across all files need to be consolidated or whether it's good enough to leave them as strings.

ZhukovAlexander:
I don't mind discarding this, unless someone else thinks that this is useful. It's just that for me it was sometimes hard to trace the origins of some strings, until I've realised that those are states. I don's know which one is simpler. Probably, we don't need this.

PR #134, opened 2016-12-02
Maintain a task tree in a kernel

ZhukovAlexander:
This PR introduces a new API for getting a list of child tasks for a particular tasks. This functionality is enabled by the change to the kernel data structure for storing tasks.

The new implementation of the kernel tasks data structure might serve as a ground up for an advanced task management, such as recursive task cancellations, for example.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -619,6 +619,9 @@ class Kernel(object):
         def _sync_trap_get_current():
             return current
 
+        def _sync_trap_get_child_tasks(task):
+            return list(tasks[task.id][2])
+
         # Return the current value of the kernel clock
         def _sync_trap_clock():
             return time_monotonic()
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -332,8 +332,12 @@ class Kernel(object):
             task.terminated = True
 
             _, parent, child_tasks = tasks[task.id]
+
+            # there are cases when parent is already gone from tasks, so
+            # we need to check if it is still there
             if parent is not None and parent.id in tasks:
                 tasks[parent.id][2].remove(task)
+            # set the maintask to be a new parent to avoid orphaned tasks
             for child in child_tasks:
                 tasks[child.id][1] = maintask
             del tasks[task.id]
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -44,7 +44,7 @@ class Kernel(object):
 
         self._selector = selector
         self._ready = kqueue()            # Tasks ready to run
-        self._tasks = { }                 # Task table
+        self._tasks = {}                 # Task table
         # Dict { signo: [ sigsets ] } of watched signals (initialized only if signals used)
         self._signal_sets = None
         self._default_signals = None      # Dict of default signal handlers
@@ -183,7 +183,7 @@ class Kernel(object):
         wake_queue = self._wake_queue           # External wake queue
 
         # ---- Number of non-daemonic tasks running
-        njobs = sum(not task.daemon for task in tasks.values())
+        njobs = sum(not task[0].daemon for task in tasks.values())
 
         # ---- Bound methods
         selector_register = selector.register
@@ -205,7 +205,7 @@ class Kernel(object):
             task = Task(_kernel_task(), taskid=0, daemon=True)
             _reschedule_task(task)
             self._kernel_task_id = task.id
-            self._tasks[task.id] = task
+            self._tasks[task.id] = [task, current, set()]
 
         # Internal task that monitors the loopback socket--allowing the kernel to
         # awake for non-I/O events.  Also processes incoming signals.  This only
@@ -253,7 +253,9 @@ class Kernel(object):
         def _new_task(coro, daemon=False):
             nonlocal njobs
             task = Task(coro, daemon)
-            tasks[task.id] = task
+            tasks[task.id] = [task, current, set()]
+            if current is not None:
+                tasks[current.id][2].add(task)
             if not daemon:
                 njobs += 1
             _reschedule_task(task)
@@ -329,6 +331,11 @@ class Kernel(object):
                 task.joining = None
             task.terminated = True
 
+            _, parent, child_tasks = tasks[task.id]
+            if parent is not None and parent.id in tasks:
+                tasks[parent.id][2].remove(task)
+            for child in child_tasks:
+                tasks[child.id][1] = maintask
             del tasks[task.id]
 
         # For "reasons" related to task scheduling, the task
@@ -347,14 +354,14 @@ class Kernel(object):
         # resources are cleaned up.
         def _shutdown():
             nonlocal njobs
-            tocancel = [task for task in tasks.values() if task.id != self._kernel_task_id]
+            tocancel = [task[0] for task in tasks.values() if task[0].id != self._kernel_task_id]
 
             # Cancel all non-daemonic tasks first
             tocancel.sort(key=lambda task: task.daemon)
 
             # Cancel the kernel loopback task last
             if self._kernel_task_id is not None:
-                tocancel.append(tasks[self._kernel_task_id])
+                tocancel.append(tasks[self._kernel_task_id][0])
 
             if tocancel:
                 _new_task(_shutdown_tasks(tocancel))
@@ -695,7 +702,7 @@ class Kernel(object):
                     # operation, was cancelled, or is no longer concerned with this
                     # sleep operation.  In that case, we do nothing
                     if taskid in tasks:
-                        task = tasks[taskid]
+                        task = tasks[taskid][0]
                         if sleep_type == 'sleep':
                             if tm == task.sleep:
                                 task.sleep = None
--- a/curio/task.py
+++ b/curio/task.py
@@ -73,6 +73,9 @@ class Task(object):
         else:
             return self.next_value
 
+    async def child_tasks(self):
+        return await _get_child_tasks(self)
+
     async def cancel(self, blocking=True):
         '''
         Cancel a task by raising a CancelledError exception.
--- a/curio/task.py
+++ b/curio/task.py
@@ -74,6 +74,9 @@ class Task(object):
             return self.next_value
 
     async def child_tasks(self):
+        """
+        Get the list of all not terminated tasks spawned by this task
+        """
         return await _get_child_tasks(self)
 
     async def cancel(self, blocking=True):
--- a/curio/traps.py
+++ b/curio/traps.py
@@ -16,7 +16,7 @@ __all__ = [
     '_join_task',
     '_wait_on_queue', '_reschedule_tasks', '_queue_reschedule_function',
     '_sigwatch', '_sigunwatch', '_sigwait', '_get_kernel', '_get_current',
-    '_set_timeout', '_unset_timeout', '_clock', 
+    '_get_child_tasks','_set_timeout', '_unset_timeout', '_clock',
     ]
 
 from types import coroutine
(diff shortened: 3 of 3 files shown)


dabeaz:
The very first versions of curio had a task tree (sort of inspired by Unix).  I thought it might be interesting to do some things like cancellation across the tree. For example, if the parent got cancelled, then all of the children would get cancelled with it.    This was implemented by maintaining a set of child tasks on the `Task` class itself.   Operations such as `spawn()` and `cancel()` would maintain this set. 

At some point, I took it out because it seemed like it was getting too complicated---and maybe it seemed better to just leave it up to an application to manage the task relationships on its own.  Or, I just wanted to see how things developed with curio to get a sense for whether or not this would be a good idea or not.

I'd actually be pretty curious to know how people feel about having tasks track their child tasks or not.  Is that a useful thing?  How would you want it to interact with things like cancellation? Daemonic tasks?

imrn:
Is the following style should be supported? :

- Spawn tasks and let their handles go into the oblivion
even if you know at a later time you need them.
- Then require the framework to provide all relations
between them.

Not exactly but there is again a smell of
"Garbage Collected" approach here.

ZhukovAlexander:
@imrn , one argument for this is that if a task crashes, if the exception is not handled, there will remain orphaned, and potentially will consume resources of kernel without any use.

imrn:
David: What exactly does Curio do on an unhandled exception in a task?

dabeaz:
Right now, an unhandled exception results in the exception information (sys.exc_info data) to be stored on the `Task` object.  The `Task` is removed from the in-kernel task table.   The associated traceback is optionally logged (True by default).   If another task tries to `join()` with the crashed task, a chained `TaskError` exception is raised.    

I'm not aware of an implicit resource leak with a crashed task.  The kernel basically lets the task go and no longer holds any references to it.  The associated `Task` instance could be held alive somewhere else (maybe code waiting for a `join()`), but otherwise it's assumed that it would be garbage collected.  Note to self: it would be good to verify this.

ZhukovAlexander:
I think this approach could bring curio up to a new level, compared to other concurrent programming libraries. Right now only curio can afford such advanced features.

dabeaz:
I think my number one concern with maintaining a task tree in-kernel is simply the mechanics of maintaining it in a sane way.   It's very easy to create small, short-lived tasks in curio for all sorts of purposes.   I'd want to make sure that's reasonably efficient.   I'm also concerned about the whole garbage collection side of it.  For example do tasks stick around until they're explicitly joined?   Are there issues with "zombie" tasks.   That kind of thing.

Definitely want to think about how all of this would work with all of the corner cases.

imrn:
Should an application keep track of its own
subprocesses, threads, connections, http requests,
open files, etc, etc, and other resources? Yes ? May be? No?

What is different with its created tasks?

Other than this, there is quiet a bit of discussion for cyclic references
involving exception leaks on python. I guess you are making a general
projection on to Curio, rather than having caught aspecific
bug. If you have a caught a specific bug please eloborate
with a 'reproducable case'.

And please note that all such evil stems from
Garbage Collection/Tracking. Your proposal adds another
layer for it.

njsmith:
My general feeling is that curio needs a real framework for managing tasks, and I don't know whether this is a useful component in that or not :-).

What I mean by a "framework" is basically, some sort of APIs or conventions, which might or might not be baked into curio, that makes it natural and ergonomic to write code where when a child task crashes that always results in the exception being propagated to somewhere useful, and when a parent task crashes then it properly cleans up child tasks. On top of that you probably want tools for e.g. respawning crashed tasks and so forth, but those are the two key points.

I think this is a pretty big thing, like: I think we probably will eventually want to redo how `run` works so to match these conventions, and have a first-class notion of "who is supervising" each task, and if an unsupervised task crashes then `run` should crash, etc. But it will take a while to get there!

As a first step I've started exploring this in my half-written HTTP server... the code isn't public yet, but here's a snapshot of the basic task supervisor code I'm working with for now: https://gist.github.com/njsmith/a655f364e61cfda94d377a2ee2f9c7cb

In my current design, the server sets up a "root" supervisor, then spawns two "child" supervisors into it: one for listener tasks, and one for per-connection tasks. The split is handy because when doing a graceful shutdown, we kill the listener supervisor (and all the tasks in it) to stop accepting new connections, and then wait for the connection supervisor to drain. (Interesting point: this means that the supervision tree actually does not match the spawn tree -- listeners spawn connections, but the connection supervisor is a sibling of the listener supervisor.) When I get around to HTTP/2 support, I expect that each HTTP/2 connection will also be its own supervisor managing a read task, a write task, and one task per active stream.

I've also been reading about Erlang "supervision trees" as a source of inspiration:
* http://erlang.org/doc/design_principles/sup_princ.html
* http://erlang.org/doc/man/supervisor.html

The ideas don't all map directly, but they're definitely useful. One interesting thing I learned from them is that when spawning a task, you want to have a concept of "initialization" or "startup", i.e., some code that runs in the context of the *parent* before the actual task spins off. So for example, my API for spawning an HTTP listener currently looks like:
```python
async def start_listener(listener_spawn_fn, ...):  
    listen_sock = await create_and_bind_listen_sock()
    try:
        set_tcp_options(listen_sock)
        await listen_sock.listen()
        return await listener_spawn_fn(
    except:
        await listen_sock.close()
        raise
```
and you call it like `await start_listener(curio.spawn, ...)` or `await start_listener(supervisor.spawn, ...)`. What this gives you is (a) if there's some error during startup (e.g. the bind fails because there's another socket there) then that failure propagates to the caller synchronously, (b) by the time we return to the caller, the socket is actually open and listening and ready to be connected to, which helps avoid a common race condition thing where you start a server but then have to put in a delay loop before trying to talk to it.

So I guess when it comes to a specific PR like this, my thought would be that it should be evaluated on the basis of: does this enable interesting experiments with higher-level task supervision APIs? Can it only be done inside curio itself? And does it interfere with code that isn't experimenting with this particular feature?

In this case, the answers might be:
* it seems like there could be interesting APIs that this enables, though it's a bit vague right now. The case would be stronger with some concrete examples of this solving a realistic problem.
* This would indeed be somewhat tricky to implement outside the curio core, though I suppose one could do something like writing a custom wrapper around `curio.spawn` that tracks the extra information, and then using that?
* My main worry about how this could affect other code is the same as @dabeaz's, that it will end up pinning task objects in memory past their life-time. In particular, it looks like in the current patch, if a parent task exits, then the corresponding `Task` object won't be cleaned up until all of its children have exited?

Regarding the code itself, I really dislike the use of the 3-item-list as an ad-hoc data structure... it makes the code confusing to read. It would be better to use a real object of some code (I like [attrs](https://attrs.readthedocs.io/en/stable/) for little container objects), or better yet, put the attributes directly on the task object.

dabeaz:
Was thinking about this some more on my morning bike ride.   There is definitely an opportunity to do something interesting here.  But, some thoughts...

On resource management and garbage collection:   

Once you start building relationships between tasks, there is the potential for things to get really tricky really fast.  For example, even writing code that implements something as "simple" as a publish-subscribe pattern has a ton of corner cases.  Pretty soon, you're up to your eyeballs in weak references and WeakKeyDicts.  Given that coroutines are already rather tricky with garbage collection (e.g., finalizing them correctly), the idea of mixing them with weak references makes me wonder whether it would spawn a zombie swarm of brain eating wasps or something.    

Right now, the Curio kernel tries to keep it real basic.  There is a dictionary of active tasks.  When a task terminates, it is removed from the task dictionary.  That is it.  There are no other Task references lurking around.  No implicit relationships get built.  No reference cycles.  Thus, it's something you can reason about.

That doesn't mean that Task instances couldn't be held elsewhere.  For example, they might be held in coroutines that are waiting to join or perform other kinds of operations.   That's outside the control of the kernel though and not the kernel's concern.  I'm okay with that.

On crash handling:

It occurs to me that the current strategy of crash handling is just broken.  Right now, Curio invokes a synchronous callback function in response to a crash.  However, there is nothing interesting that can be done there.   You can't interact with the kernel (there's no synchronous API). You can't invoke coroutines.  You can't respawn a replacement task.  The only thing you can do is maybe log an error message.  That's just dumb.  We can do better.

On tracking child tasks:

Outside of debugging, I wonder about the utility of a task knowing the tree of all descendant tasks.  For example, it seems that if a parent task is spawning children, it is already quite easy for it to keep track of them independently of the kernel:

          children = []
          ...
          t = await spawn(child())
          children.append(t)

If one of those child tasks spawns another task, is that something that the parent would want to know about?  I just don't know.   For example, suppose a child runs the `curio.subprocess.check_output()` coroutine.  Under the covers, that operation spawns two subtasks as part of its internal operation.  Is that something that needs to be tracked or is it an incidental detail?   What would a parent task do with that knowledge?   It seems that tracking and doing anything interesting with implicitly created tasks like that would be hard and fraught with problems.    

On improving crash handling and task exit:

It occurs to me that there are some interesting other options for handling crashes and exit cases.  One idea would be to give curio tasks some form of an `atexit()` function.   Maybe something roughly like this:

    children = []
    ...
    t = await spawn(child())
    children.append(t)
    await t.atexit(children.remove, t)

(could be some interesting facets of this if `atexit` accepts both synchronous and async functions as a callback).

Another idea is to have the kernel optionally push all crashed tasks (and maybe terminated tasks) onto an optionally provided queue.  For example, maybe you do this:

    async def done_handler(done_queue):
            while True:
                   task = await done_queue.get()
                   ... do something

    async def parent():
            done_queue = curio.Queue()
            await spawn(done_handler(done_queue))
            ...
            t = await spawn(child(), exit_queue=done_queue)
     
The gist of the idea is that you make some kind of "supervisor" task that watches the queue.  When tasks terminate, it is notified and can take action.  Since the supervisor is a proper coroutine, it could take any number of more interesting steps such as restarting a task.

I kind of like this queue idea.   It seems like a lot of interesting things could be done if the kernel merely pushed terminated tasks onto an optionally specified user queue.  Plus, it's super easy to implement kernel-side. 

Anyways, those are my random thoughts for the moment.

kdart:
For my project I wrote my own process manager based on curio. I'd prefer to keep curio simple and get bloated like asyncio. I like having the flexibility of using other modules, or my own, to handle these other various functions.

ZhukovAlexander:
@dabeaz , I also like the idea of a "trash queue" for tasks. 

As for the task tree, this might also be useful for the "monitor" feature. I can imagine a case, when some evil task spawns a lot of new tasks, that will DOS the event loop in some way. Obviously, if you use curio monitor to manually kill all those tasks, you'll be out of luck, since cancelling say 100 child tasks is already hard, while this could be done pretty smart if we implement a "group cancellation" on top of this PR. But again, I'm just theory crafting. I would also prefer to keep curio as simple and robust as possible, though I don't know whether this particular patch is harmful. Let's see what the community thinks.

>  if a parent task exits, then the corresponding Task object won't be cleaned up until all of its children have exited?

@njsmith , actually, when a parent is cleaned up, all it's child will have a main task as their new parent, just like in unix all child processes become child of init process when a parent exits.

asvetlov:
Hmm. In Unix we have zombies, Linux has tricks to acquire ownership on orphaned sub-sub-processes etc. and cgroups at the end..
Process tree support is really very complex problem with many gotchas.

njsmith:
The enqueue-task-on-exit model makes a lot of sense to me too. It's simple and surely not the end of the story, but it formalizes something that my supervisor and `curio.task.wait` are already doing in somewhat more awkward ways (so that's good evidence that it's practically useful), and it at least provides an answer to the current annoying situation where the kernel has no way to know which task crashes are actually an unexpected event that needs to trigger noise on the console, and which ones are fine because there's a supervisor that will catch and handle the exception.

imrn:
Take my +1 for "Seperate task manager app" approach which would not
penalize anyone who never needs such a functionality. Better crash handling
may seem fine at first sight. However things that should be done on a crash
is completely application dependent.

Even if curio crash handling is improved "what will it do?" Restart??
Definetely not a good idea.. Your use case declares clean up, restart,
continuation, termination conditions.

I always write Python code as if it is C++ with complete control
on the resources it manages. No general cleverness exists to
deal with messy coding.. If you can't trust your code then
write "try:" on the very first line of your coroutine.
And spend some time developping a good task hieararcy
with bullet proof task management strategy specific to your
application. Nobody can give it to you..

It is interesting that many of the requests here are of this type.

ZhukovAlexander:
I think I'd close it.

PR #135, opened 2016-12-03
Allow personalization of debugger used.

kdart:
Also change the flag to "debug" to be more readable.

I prefer using a different debugger than pdb. This is a simple change allows an environment setting to choose an alternate debugger.


dabeaz:
I've removed debugger functionality entirely from the kernel for the moment. Rethinking crash handling .

PR #139, opened 2016-12-08
Cache current loop cycle clock value

ZhukovAlexander:
This PR is an implementation of the [libuv smart time caching](http://docs.libuv.org/en/v1.x/design.html#the-i-o-loop). This has the following benefits:

- all traps, that rely on a kernel's clock will have a slight performance increase
- avoid multiple calls to `time.monotonic` during a single event loop ineration
- all tasks that are being run during the same loop iteration will see the same value of the kernel's clock, which is what happens in a truly parallel environment.

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -191,6 +191,7 @@ class Kernel(object):
         ready_popleft = ready.popleft
         ready_append = ready.append
         time_monotonic = time.monotonic
+        now = time_monotonic()
         _wake = self._wake     
 
         # ---- In-kernel task used for processing signals and futures
@@ -510,7 +511,7 @@ class Kernel(object):
                 raise CancelledError("CancelledError")
             else:
                 if current.timeout is not None:
-                    now = time_monotonic()
+                    # now = time_monotonic()
                     if current.timeout <= now:
                         raise TaskTimeout(now)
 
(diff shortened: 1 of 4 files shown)


njsmith:
Interesting idea!

Since part of the motivation here is presumably speed, have you measured
the performance impact?

I find the part where the public clock() trap always returns the same value
to be quite surprising, and the part where relative timeouts use the stale
time when converting to absolute times to be somewhat surprising. Does
libuv do this too?

dabeaz:
I'm kind of surprised by `clock()` returning the same value as well.   What happens if the ready queue fills up with a bunch of stuff that takes a bit of time to work through and one of the later tasks calls `clock()`.  It seems that its time value could be skewed in some way.   

I'm thinking that `clock()` should return the most accurate time possible.   Partly because certain parts of curio have been written in a way to allow for periodic heartbeat timers and things like that (e.g., having a task wake up at precise periodic intervals).   It seems that knowing the time as precisely as possible would be essential for that.

tahajahangir:
Note that `clock` and `gettimeofday` system calls actually are virtual system calls, and are pretty fast (even with CPython wrappers).

```
In [4]: def f(): pass

In [5]: %timeit f()
The slowest run took 34.07 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 90.8 ns per loop

In [6]: %timeit time.monotonic()
The slowest run took 25.96 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 115 ns per loop

In [7]: t = time.monotonic

In [8]: %timeit t()
The slowest run took 31.04 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 81.3 ns per loop

In [9]: i = 0

In [10]: %timeit i + 1
The slowest run took 63.23 times longer than the fastest. This could mean that an intermediate result is being cached.
10000000 loops, best of 3: 43 ns per loop
```
Calling `time.monotonic` is cheaper than calling an empty python function!

PR #144, opened 2016-12-21
PEP8-ify code base (round 1/3)

giampaolo:
This PR fixes most PEP8-related errors. Amongst others:
- removal of unused imports
- lines len max 79 chars
- 2 lines between functions 

Before the patch:

```
~/svn/curio {master}$ make flake8  | wc -l
558
```

After the patch:

```
~/svn/curio_giampaolo {pep8ify_code}$ make flake8  | wc -l
107
```

The remaining 107 issues are all related to `from x import *` statements which I will handle separately in another PR because they break a lot of things.

If desired I will later provide a third and final PR to add a GIT hook which rejects un-compliant code on commit and include `make flake8` as part of the CI process on Travis.


ZhukovAlexander:
I would ignore max line length. It's usually being set to a higher limit instead of 79 characters, because it is pretty annoying to maintain. Just an opinion though.

giampaolo:
I think it's more common to find projects sticking with 79 chars than the other way around, as that's what PEP-8 dictates. As a consequence 79 is the default in many linters such as `pep8` and `flake8` which are used by most editors and IDEs in order to provide real time syntax-highlighting of uncompliant lines. For instance, Sublime assumes 79 by default and probably also PyCharm and Eclipse.

njsmith:
I personally use a 79 char limit in my code, but it's much more important that a project have some internal consistency that matches how the project wants to work than that it match PEP 8 (see the very first rule in PEP 8). kernel.py in particular is obviously intentionally not written to a 79 character limit... personally I guess I have a mild preference that it would be, but I'm not going to argue about it :-). @dabeaz should just lay down the law on that kind of thing IMO.

Most of this patch seems to be cleaning up genuinely screwy stuff like 3-space or 5-space indents, function calls like with inconsistent spacing like `f(1, 2,3)`, etc., which definitely seems worth cleaning up.

Personally I find automatic-flake8 rules maddening and never use them on my projects because 100% of the time that they trigger for me, it's forcing me to uglify my code. I guess I am more consistent about manually typing things in correct style than some though...

dabeaz:
As a general comment, I am not a stickler on rigidly enforcing PEP-8 style.    I don't mind patches that try to clean things up, but would not want the project to have a strict requirement of code always having to conform to PEP-8 for consideration of a pull request.  As such, I do not want any kind of git commit hooks that reject code. 

I would also ignore max line length.   Curio uses a lot of nested functions and I don't want to be constrained in that way.  I'm not suggesting that lines would expand to crazy lengths, but I really don't have an issue with 100 character lines.

dabeaz:
Just quickly reviewing the patched files, there are too many lines that are now wrapped in a funny way (for my tastes).  Is there any way to redo this without a line limit?

giampaolo:
OK, fair enough. 

 > Is there any way to redo this without a line limit?

Yup. I'll do that.

giampaolo:
> As such, I do not want any kind of git commit hooks that reject code.

How would you feel about a GIT hook which uses `pyflakes` instead of `pep8` / `flake8` which rejects "bad behaviors" (not bad styles)? For instance, this:

```
git ls-files | grep \\.py$ | xargs $(PYTHON) -m pyflakes
```

...executed against the current code complains about a lot of undefined variable names due to `from x import *` and is able to catch variable name typos, invalid syntax etc., e.g.:

```
~/svn/curio_giampaolo {pep8ify}$ git ls-files | grep \\.py$ | xargs python3.5 -m pyflakes | grep -v undefined
curio/socket.py:14: 'socket.*' imported but unused
curio/ssl.py:11: 'ssl.*' imported but unused
examples/bench/curiostream.py:1: 'curio.spawn' imported but unused
tests/test_io.py:33: local variable 'data' is assigned to but never used
tests/test_io.py:75: local variable 'data' is assigned to but never used
tests/test_io.py:117: local variable 'data' is assigned to but never used
tests/test_kernel.py:454: local variable 'task' is assigned to but never used
tests/test_kernel.py:750: local variable 'results' is assigned to but never used
tests/test_queue.py:3: 'collections.deque' imported but unused
tests/test_queue.py:138: local variable 'item' is assigned to but never used
tests/test_queue.py:188: local variable 'item' is assigned to but never used
tests/test_signal.py:3: 'time' imported but unused
tests/test_signal.py:20: local variable 'task' is assigned to but never used
tests/test_signal.py:51: local variable 'task' is assigned to but never used
tests/test_socket.py:3: 'os' imported but unused
tests/test_socket.py:238: local variable 'data' is assigned to but never used
tests/test_socket.py:247: local variable 'task' is assigned to but never used
tests/test_socket.py:280: local variable 'data' is assigned to but never used
tests/test_subprocess.py:5: 'curio.*' imported but unused
tests/test_subprocess.py:48: local variable 'out' is assigned to but never used
tests/test_subprocess.py:61: local variable 'out' is assigned to but never used
tests/test_subprocess.py:76: local variable 'out' is assigned to but never used
tests/test_sync.py:114: local variable 'task' is assigned to but never used
tests/test_sync.py:147: local variable 'task' is assigned to but never used
tests/test_sync.py:630: local variable 'task' is assigned to but never used
tests/test_workers.py:109: local variable 'result' is assigned to but never used
tests/test_workers.py:129: local variable 'result' is assigned to but never used
```

Would you OK with something like this and / or getting rid of `from x import *`

dabeaz:
I don't want any automatic git hooks installed that would reject code.  I would prefer to handle these issues manually through a makefile or some other process.

On `from x import *`, that is used for very specific reasons in certain places.  For example, the `socket` and `ssl` modules do that because they are meant to operate as stand-ins for the builtin modules of the same name.  As such, they need to export all of the predefined constants and other symbols that are defined in those modules.   This style of import is also used in `__init__.py` files where is it well controlled through the use of `__all__` variables.

giampaolo:
I understand the rationale behind using wildcard import in `ssl` and `socket` modules. I would just avoid it elsewhere.

giampaolo:
I created https://github.com/dabeaz/curio/pull/145 which in which I assume max 100 chars per-line. Closing this one out.

Issue #83, opened 2016-10-15
Maybe curio should wait for socket being writeable before attempting to write?


njsmith:
So I just stumbled across a rather arcane corner of the Linux/OS X socket API while trying to understand why I'm seeing weird buffering behavior in a program using curio. (Actually, it's a toy program I'm using for a blog post I'm working on contrasting curio and asyncio, and one of the points I want to make is that asyncio adds superfluous user-space buffering :-). But my program with curio was seeing more buffering than I expected at the kernel level and now I'm deep down the rabbit hole.)

As we know, calling `socket.send` doesn't immediately dump data onto the network; instead the kernel just sticks it into the socket's send buffer, to be trickled out to the network as and when possible. Or, if the send buffer is full, then the kernel will reject your data and tell you to try again later. (Assuming non-blocking mode of course.)

But for various reasons, it turns out that the kernel's send buffer is usually way larger than you actually want it to be, which means you can end up queueing up a huge amount of data that will take forever to trickle out, introducing latency and [causing various problems](https://insouciant.org/tech/prioritization-only-works-when-theres-pending-data-to-prioritize/).

So at least Linux and OS X have introduced the euphoniously named and [terribly documented](https://lwn.net/Articles/560082/) `TCP_NOTSENT_LOWAT` feature. Basically what it does is let you use `setsockopt` to tell the kernel -- hey, I know that you're *willing* to buffer, like, 5 MiB of data on this socket. But I don't actually want to do that. Can you only wake me up when the amount of data that's actually buffered drops below, like, 128 KiB, and I'll just top it up to there? (This is a bit of a simplification because there are some subtleties about how you budget for data that's queued to send vs. data that's been sent-but-not-yet-acked, but it's good enough to go on.)

But it turns out that TCP_NOTSENT_LOWAT *only* affects polling-for-writability. So you absolutely can have a socket where `select` and friends say "not writeable", but *at the same time* `send` is happy to let you write lots and lots of data. And this is bad, because it turns out literally the only way the kernel is willing to give you the information you need to avoid over-buffering is with that "not writeable" signal.

So if you want to over over-buffering, then you have to always call `select`-or-whatever before you call `send`, and only proceed if the socket is claimed to be writeable.

And unfortunately, right now, curio never does this: it always tries calling `send` first, and then only if that fails does it block waiting for writeability.

The solution is simple: before calling `socket.send`, check that the kernel thinks the socket is writeable.

[...]

PR #146, opened 2016-12-22
Use poll() instead of select() in writeable()

giampaolo:
...in order to avoid crashing if `fd > 1024`.


giampaolo:
Also references https://github.com/dabeaz/curio/commit/7f7a680fa815fe12d83eab6efc2e1813a93b1008 and https://github.com/dabeaz/curio/issues/83.

njsmith:
So the fallback here continues to crash if the `fd` is too large and `PollSelector` isn't available (I think this might happen on Windows?). Since this is just an optimization in the first place, maybe a better fallback would be to go directly to `_write_wait`? Or possibly even remove the optimization entirely, I guess. IIRC I suggested it in the first place but I don't think it's very important :-)

giampaolo:
Right, `_write_wait()` is probably a better fallback than select(). =)
To be honest now that I look at this I don't understand the rationale behind `writeable`. Some questions:

- I'm probably missing something but given the buffer issue described in #83, shouldn't this (checking if fd is writeable) be done in all `send*` methods *before* the actual `socket.send*` call?
- I noticed `writeable()` is not documented and not used  anywhere except in a unit test. What's the use case of `writeable()` from a user standpoint?

njsmith:
Have you read through #83? Both of those are explained downthread (though I know there's a lot of text there...)

njsmith:
https://github.com/dabeaz/curio/issues/83#issuecomment-254103790 in particular might be relevant, but probably it needs context to make sense...

dabeaz:
Looking at this, I'm thinking of going straight to _write_wait() as well.  Really, the point of the writeable() method was to provide a way to force I/O polling instead of eagerly trying to send if that's what you really needed to do.    It's probably fine to enter the kernel for that.

giampaolo:
Done.

JordanP:
I am following closely the git history of Curio. This PR was a bit messy. Please squash your changes before opening a PR, especially when your intermediary commits are not really meaningful. Thanks.

giampaolo:
Yeah it was a bit messy indeed. Sorry about that.

PR #155, opened 2017-01-11
Fix the name of a kwarg to the `abide` function

JordanP:
The kwarg is named `reserve_thread` not `reserve`, so use
`reserve_thread` consistently everywhere.

### DIFF
--- a/curio/sync.py
+++ b/curio/sync.py
@@ -326,7 +326,7 @@ def abide(op, *args, **kwargs):
     methods execute in a background thread.  Cancellation is handled
     gracefully. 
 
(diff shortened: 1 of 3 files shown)


JordanP:
I am not wrong with this PR, it makes me wonder whether `test_abide_sync_rlock()` actually tests what it should test.

dabeaz:
I don't know how to make Curio support re-entrancy on a foreign RLock (well, at least not easily).  So, that test is currently only testing to see whether it can synchronize with the lock at all--under the constraint that the same thread must be used to acquire and release the lock.

PR #159, opened 2017-01-20
Don't try and select until after ready events have been processed

### LINKED ISSUE #75: Windows 10, OSError: [WinError 10022] An invalid argument was supplied
Hi!
Running the following ["getting started"](https://curio.readthedocs.io/en/latest/tutorial.html#getting-started) code:

``` python
async def countdown(n):
    while n > 0:
        print('T-minus', n)
        await curio.sleep(1)
        n -= 1

if __name__ == '__main__':
    curio.run(countdown(10))
```

results in the following output:

```
T-minus 10
Traceback (most recent call last):
  File ".../Projects/curio_test/test.py", line 24, in <module>
    curio.run(countdown(10))
  File "...\VirtualEnvs\curio_test\lib\site-packages\curio\kernel.py", line 613, in run
    result = kernel.run(coro, shutdown=True)
  File "...\VirtualEnvs\curio_test\lib\site-packages\curio\kernel.py", line 491, in run
    events = selector_select(timeout)
  File "...\Python\Python35\lib\selectors.py", line 323, in select
    r, w, _ = self._select(self._readers, self._writers, [], timeout)
  File "...\Python\Python35\lib\selectors.py", line 314, in _select
    r, w, x = select.select(r, w, w, timeout)
OSError: [WinError 10022] An invalid argument was supplied
```

OS: Windows 10 x64
Python: 3.5.2

swfiua:
Previous attempts to solve #75 included moving _init_loopback_task outside the main run() loop.

This fails to solve the problem because new tasks are simply added to the *ready* queue and that is not processed until the selector is called to look for new events.

This fixes #75 by moving the selector code and subsequent processing of events to the end of the run loop.

It does indeed solve the problem on windows, so far without an increase in stinging bat density.

It should be noted I'm quite out of my depth at this point, so proceed with caution!

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -717,45 +717,6 @@ class Kernel(object):
                 timeout = sleeping[0][0] - time_monotonic()
             else:
                 timeout = None
-            events = selector_select(timeout)
-
-            # Reschedule tasks with completed I/O
-            for key, mask in events:
-                rtask, wtask = key.data
-                intfd = isinstance(key.fileobj, int)
-                if mask & EVENT_READ:
-                    # Discussion: If the associated fileobj is
-                    # *not* a bare integer file descriptor, we
-                    # keep a record of the last I/O event in
-                    # _last_io and leave the task registered on
-                    # the event loop.  If it performs the same I/O
-                    # operation again, it will get a speed boost
-                    # from not having to re-register its
-                    # event. However, it's not safe to use this
-                    # optimization with bare integer fds.  These
-                    # fds often get reused and there is a
-                    # possibility that a fd will get closed and
-                    # reopened on a different resource without it
-                    # being detected by the kernel.  For that case,
-                    # its critical that we not leave the fd on the
-                    # event loop.
-                    rtask._last_io = None if intfd else (key.fileobj, EVENT_READ)
-                    _reschedule_task(rtask)
-                    mask &= ~EVENT_READ
-                    rtask = None
-                if mask & EVENT_WRITE:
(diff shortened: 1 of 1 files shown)


dabeaz:
Honestly not sure how I feel about moving that code to the end.   Almost wondering if we can fix this with some kind of simple check. Or maybe even just an exception handler (catch the offending exception when it's empty and move on).

swfiua:
The way I look at this, it is an infinite loop, so the fix only really has an effect on the first time round the loop -- it is an infinite cycle of "..  events, sleepers, ready ... ".

So the only effect of moving the code is that sleepers and ready stuff now get processed the first time round the loop, before the selector gets called, whereas in the past selectors were done first.

Now, because of the way curio works, the first time round the loop (on unix systems) an empty list of events will be returned, so the first iteration of the loop no events are processed anyway.

So the fix doesn't really change anything except avoiding calling selector_select when the selector is empty.

Plan B would be to do something like this at the top of the loop:

```python

   if selector._readers or selector._writers:
        events = selector_select(timeout)
  else:
        events = []

Note that this has the same (practical) result as moving the selector code to the end of the loop, since the first time round the loop the selector is always empty, so no event code gets executed anyway until the sleepers and ready stuff have been processed the first time round.

swfiua:
This version just catches the OSError exception from selector_select call with nothing to select.

dabeaz:
Going to merge this.  I had a thought though.  Instead of logging an error message, what if the exception handler merely slept for the specified timeout period?  With that, maybe the loopback task wouldn't be necessary.

PR #160, opened 2017-01-21
make curio.local thread-safe

dutc:
When managing task-local storage, make sure the task-local storage is
connected to the current executing thread, allowing the creation of
multiple curio kernels, one-per-thread.

### DIFF
--- a/curio/local.py
+++ b/curio/local.py
@@ -60,7 +60,6 @@ from contextlib import contextmanager
 # The thread-local storage slot that points to the task-local storage dict for
 # whatever task is currently running.
 _current_task_local_storage = threading.local()
-_current_task_local_storage.value = None
 
 
 @contextmanager
@@ -68,7 +67,11 @@ def _enable_tasklocal_for(task):
     # Using a full save/restore pattern here is a little paranoid, but
     # safe. Even if someone does something silly like calling curio.run() from
     # inside a curio coroutine.
-    old = _current_task_local_storage.value
+    try:
+        old = _current_task_local_storage.value
+    except AttributeError:
+        old = None
+
     try:
         _current_task_local_storage.value = task.task_local_storage
         yield
--- a/tests/test_local.py
+++ b/tests/test_local.py
@@ -2,6 +2,7 @@
 
 import pytest
 from curio import *
+from threading import Thread
 
 # Like run, but unwraps exceptions so pytest can see them properly.
 # Lets us use assert from inside async functions.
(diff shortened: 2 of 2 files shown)


dutc:
This was driven by a stupid example (https://gist.github.com/dutc/5bb2dee142f037876efadee4e99c134c) showing the use of `curio.Task` from within a `threading.Thread` from within a `multiprocessing.Process`.

curio/local.py
```diff
@@ -58,22 +58,28 @@
...
-_current_task_local_storage.value = None
+# whatever task is currently running in whichever thread is currently running.
+_thread_local_storages = {}
 
+def _local_storage():
+    thread = threading.current_thread()
+    if thread not in _thread_local_storages:
+        _thread_local_storages[thread] = threading.local()
+        _thread_local_storages[thread].value = None
+    return _thread_local_storages[thread]
```
njsmith: So uh... this is exactly a reimplementation of what `threading.local` does (only `threading.local` is more efficient because it can advantage of the platform native mechanisms).

njsmith:
Do you have an example of what the problem was that motivated this? Does the test given there actually fail on curio mainline? Because AFAICT the code changes here are a no-op :-)

dutc:
The test fails without these changes.

There is no actual use-case associated with this patch.

The current code fails, because the `.value` attribute is created only on the `PyThreadState.dict` for the thread which imported the module itself.

Here's a simpler patch.

curio/local.py
```diff
@@ -60,15 +60,18 @@
...
 @contextmanager
 def _enable_tasklocal_for(task):
     # Using a full save/restore pattern here is a little paranoid, but
     # safe. Even if someone does something silly like calling curio.run() from
     # inside a curio coroutine.
-    old = _current_task_local_storage.value
+    try:
+        old = _current_task_local_storage.value
+    except AttributeError:
+        old = _current_task_local_storage.value = None
```
njsmith: doh, this is a silly bug indeed, good catch :-). You could make the patch slightly simpler even by leaving out this assignment to `.value`.

dutc:
How about that?

njsmith:
Looks good to me! I don't have merge permissions though, so it will have to wait for @dabeaz to stop by.

Also, I feel silly that I only just realized "dutc" = dontusethiscode. hi James!

dabeaz:
Going to merge, but test might need tweaking for Python 3.5 compatibility (if we're keeping that ;-).

Issue #111, opened 2016-11-15
curio and tkinter


nameoftherose:
Thanks for curio and the youtube tutorials. They are great.
I have a number of sensors sending their measurements over the LAN. Using curio I collect all measurements.
I now need to integrate tkinter to the program.
I would be thankfull for any suggestions.
Thank you.

nameoftherose:
[Neat discovery: How to combine asyncio and tkinter : Python - Reddit](https://www.google.gr/url?sa=t&rct=j&q=&esrc=s&source=web&cd=1&cad=rja&uact=8&ved=0ahUKEwiSvr3k_q3QAhXFCBoKHQ9dAf0QFgggMAA&url=https%3A%2F%2Fwww.reddit.com%2Fr%2FPython%2Fcomments%2F33ecpl%2Fneat_discovery_how_to_combine_asyncio_and_tkinter%2F&usg=AFQjCNHGx5vWGlX0PLFj5HmkvqQQY0vwhw) shows a way to do this using the asyncio module. The concept translates in a straightforward way to **curio** with one exception: in order to spawn a new coroutine in response to a gui event, it uses the asyncio.async function (equivalently the asyncio.ensure_future). 
It seems there is no such functionality in curio. (curio.spawn is an async def function, cannot be called from a tkinter callback). 
Could such a function be added?  Or could you give some advice so I try to add it myself?
Thank you.

dabeaz:
That is pretty interesting.  I have a solution! (I think).   The solution is to use a curio queue object.  Create a queue and a task that is solely dedicated to spawning new tasks::

     import curio
     newtask_queue = curio.Queue()

     async def creator_task():
              while True:
                      coro = await newtask_queue.get()
                      await curio.spawn(coro)

Launch this task when your curio application starts::

    async def main():
         await curio.spawn(creator_task())
         ...

Now, in your Tk callbacks, use the `newtask_queue.put()` method.   For example::

     async def some_task(args):
             ...

     def callback():
             ...
             newtask_queue.put(some_task(args))
              ...

This works because the `Queue.put()` method works from normal synchronous code.   The special `creator_task()` coroutine will pick it up and spawn the new task when control makes it back to the curio event loop. 

Note:  This part of curio is pretty experimental.  Let me know if this works... ;-).

dabeaz:
[...]

PR #163, opened 2017-01-28
The tail of the OS that didn't bark in the night

swfiua:
I am just starting writing some docs around the selector_select fun and games.

The stack traces have been annoying me and I really don't like catching exceptions.  I mean, what are you going to do with them.

But if the exception is OSError ... then I probably don't want to drop them all the time.

Anyway, I'll finish the story and then we can write some code to fix this up a bit.

more to come...


dabeaz:
Yes.  Keep writing!  I want to learn.

swfiua:
I just added some more.

I'm going to fix the code to include this *selector patience* idea.

swfiua:
I haven't tested the latest fix on windows.  

There is some irony here, the next couple of weeks I'm probably not going to be using operating systems with weird selectors, but I will probably get a chance to try it on Monday.

swfiua:
This is still a work in progress.

Some more changes to the selector exception handling.

I've added a piece on event loops, my thoughts on the tkinter issue.

Also discovered today that examples/prodcons.py on Windows raises three OSError's from the selector before it settles down.  Guessing there are two curio.run's and one of them is getting to the selector really early.

So for now, I would be patient with the SELECT_PATIENCE for now.

May add a piece on platforms: what works where, what's weird.

dabeaz:
A question: should I leave all of this in a separate branch for now or should I pull it into master?   (Just looking a the "work in progress" comment above).

swfiua:
Yes, leave it here for now.

I should probably create a branch at my end, maybe the 'patience' branch.

It might be good to have the docs in master, so I will then tidy the code for this pull request to be just the doc fixes.

swfiua:
ps travis seems to be failing to get off the ground testing this branch.

swfiua:
This is a docs only change now.

Three pieces.

One on the selectors, one on event loops and a piece on platfrom quirks and quarks.

The event loop piece is me trying to understand this thread https://github.com/dabeaz/curio/issues/111 and trying to explain what I am trying to do with the UI event loops.

All very much work in progress.

dabeaz:
Closing (out of date).

PR #170, opened 2017-02-03
Make Kernel.run() return coroutine's return value

clchiou:
[Reference](https://github.com/dabeaz/curio/blob/master/docs/reference.rst) specifies that `Kernel.run()` should "run the coroutine
*coro* to completion and return its final return value."


dabeaz:
Good catch!

PR #188, opened 2017-02-27
Add basic curio -> asyncio coroutine bridge.

SunDwarf:
This allows you to run asyncio coroutines from within curio, using a new function `curio.acb` (Asyncio Curio Bridge). 

I put this inside the kernel because it's the easiest and most user-friendly way of co-operating the two event loop types.

The `asyncio` event loop runs inside a thread that is booted when `curio.run()` starts, and stops when it exits. asyncio coroutines are submitted to the worker event loop via `run_coroutine_threadsafe` and the resulting future is then awaited on. 

Right now, there's no additional tests or documentation.

### DIFF
--- a/curio/__init__.py
+++ b/curio/__init__.py
@@ -13,6 +13,7 @@ from .network import *
 from .file import *
 from .local import *
 from .channel import *
+from .bridge import *
 
 __all__ = [*errors.__all__,
            *task.__all__,
@@ -25,4 +26,5 @@ __all__ = [*errors.__all__,
            *file.__all__,
            *local.__all__,
            *channel.__all__,
+           *bridge.__all__,
            ]
--- a/curio/bridge.py
+++ b/curio/bridge.py
@@ -1,0 +1,32 @@
+# bridge.py
+#
+# Support for running asyncio coroutines from within curio.
+# The curio->asyncio bridge runs a separate asyncio event loop in a different thread,
+# which has coroutines submitted to it over the course of the kernel's lifetime.
+
+import asyncio
+
+from .traps import _get_kernel
+from .sync import Event, abide
+
+
+async def acb(coro):
+    '''
+    Runs a coroutine in the current event loop running alongside this kernel.
+    '''
+    kernel = await _get_kernel()
+    finished_ev = Event()
+    # How this works:
+    # First, it schedules a new coroutine on the kernel's asyncio loop,
+    # which will be running alongside.
+    # Then, it will simply wait for the result in another thread (yay threading!)
+    #
+    # Possible improvements:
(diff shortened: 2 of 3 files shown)


dabeaz:
Interesting.   How hard do you think it would be to make this somewhat more like thread/process pools?   For a thread-pool, there is a `thread_pool` attribute on the Kernel that only gets initialized if work is submitted to a thread.  The thread pool then gets shutdown when the kernel goes away.    So, the main difference is that the thread-setup and initialization would get pulled into the `bridge.py` file.

Actually, I'm now almost wondering if this thread pool, process pool, asyncio bridge stuff could be generalized into a more general kernel resource concept. 

I'm not going to merge this right away, but let me fiddle around with this.   Would love to get other's thoughts on it.

SunDwarf:
I did have the original idea of making this work within the threadpool rather than a single thread, but I wasn't sure how to work this properly with asyncio's threadsafety requirements. I'll tinker and see if I can make it run inside the `thread_pool`.

dabeaz:
I'm not sure it needs to run inside thread_pool.  I was thinking more about it being initialized/shutdown in a manner similar to how thread pools do it.     It could still be a separate thing.

SunDwarf:
Oh, okay. This would be possible with a small trivial change. However, there is an impl detail about that: when should the worker loop stop running? Once the asyncio coro has finished running, when the `Kernel.run` has finished, or when the kernel has shutdown fully? Right now I've made it explicit about happening when the `run()` call has finished instead of at shutdown so that the two event loops are running concurrently but independantly rather than having asyncio as a "separate" thing.

dabeaz:
I almost wonder if this could be implemented using one of Curio's new-fangled AsyncThread objects.   Hopefully you don't mind me thinking out loud about it ;-).

SunDwarf:
The `acb` function could be easily put inside an async thread - the only other await function then would be the waiting on the kernel trap. Then the `fut.result()` can be simply called.

jkbbwr:
We are hitting issues at work were we want to do something async but all the existing infrastructure is asyncio only and this is pinching slightly. Something like this could improve the situation. 

Something that should be importantly considered is the interaction of using curio queues from asyncio code.

dabeaz:
Curious to know a bit more about the "asyncio" functionality you're trying to use.   Admittedly, it's probably a bit insane to release a totally new and incompatible I/O library like Curio into the world.  I've always considered it to be a bit of a long-play that would take some time to get up to speed on supporting different kinds of services and protocols.

dabeaz:
I'm going to merge this, but may refactor it into using an async thread (which I think might be cleaner).

dabeaz:
I've opened Issue #190 for further discussion of this feature as it evolves,

PR #191, opened 2017-03-02
Prototype Promise class for discussion.

alexchamberlain:
Simple `Promise` prototype, with basic get/set and exception handling.

Open questions include:
- Do we need separate `_data` and `_exception` members, or would testing for an exception be sufficient?
- Do we need a `_done` or is `_data is not None or _exception is not None` sufficient? I think we do so you can pass back `None`.

ToDo
- Add some documentation

Thoughts? Any other test cases needed?

### DIFF
--- a/curio/__init__.py
+++ b/curio/__init__.py
@@ -14,6 +14,7 @@ from .file import *
 from .local import *
 from .channel import *
 from .bridge import *
+from .promise import *
 
 __all__ = [*errors.__all__,
            *task.__all__,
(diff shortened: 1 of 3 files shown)


njsmith:
It looks like you're reimplementing `curio.Event` along the way here?

alexchamberlain:
It is very similar, indeed. Do they serve the same purpose? If so, maybe it would be sufficient to add data/exception elements to `Event`.

njsmith:
I don't think it makes sense to add data/exception elements to `Event`, but an `Event` could replace your `done` bool + condition variable. (You'd use the condition variable's locking, but it isn't doing anything here anyway, because you never hold the lock over a yield.)

alexchamberlain:
Done; I've left it as 2 commits, but can squash before this is landed.

PR #223, opened 2017-07-02
asyncio bridge improvements

SunDwarf:
This adds some improvements to the asyncio bridge, such as:

 - A decorator that allows easier `await`ing of foreign coroutines
 - The ability to pass keyword arguments to a corofunc
 - The ability to pass a coroutine object straight into the loop.

The first one was my main use case for one of my projects; it's easier than running the loop manually every time.

### DIFF
--- a/curio/bridge.py
+++ b/curio/bridge.py
@@ -10,6 +10,7 @@ __all__ = [ 'AsyncioLoop', 'asyncio_coroutine' ]
 
 import asyncio
 import functools
+import inspect
 import threading
 
 # -- Curio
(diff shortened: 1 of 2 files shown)


dabeaz:
I'm going to merge this, but some changes will be made to the *args, **kwargs calling convention.   Throughout Curio, there has been effort to standardize the calling convention of functions that run other functions (e.g., `spawn(func, *args)`, `run_in_thread(func, *args)`, etc.).  Positional arguments are passed to the function itself whereas keyword arguments are used for configuration and options to the operation itself.   To pass keyword arguments, you would use `partial()`.

PR #230, opened 2017-09-19
Fix "Curo" typo.

donspaulding:


dabeaz:
Hmmm. "Curo".  It almost sounds like a tasty snack! Almost.

PR #254, opened 2018-04-10
Add contextvars support.

Fuyukai:
This adds support for PEP 567 contextvars, by attaching a context object to each Task and making the coroutine inside run within the context.

Problem: contextvars is only in the stdlib in Python 3.7; there is a backport for Python 3.5 and Python 3.6 but I don't know if you want to depend on that (given that curio has no dependencies).

### DIFF
--- a/curio/task.py
+++ b/curio/task.py
@@ -10,6 +10,8 @@ from collections import deque
 import linecache
 import traceback
 import os.path
+from functools import partial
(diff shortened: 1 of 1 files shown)


dabeaz:
Reviewing this.   I will probably rely on curio's Activation feature to make a version of this that is optional and doesn't impact code running on Python 3.6.

PR #257, opened 2018-04-24
fix zmq example for version 17.0.0+

dbuse:
pyzmq no longer ships nor needs its own selector.  This breaks some of the examples provided with curio.  However, the fix is easy: remove everything that relies on ZMQSelector.

This probably makes integration with other types of sockets easier (or even possible).


dabeaz:
Cool!

PR #276, opened 2018-09-13
Add contextvar support.

Fuyukai:
Should work as expected on <3.7.

### DIFF
--- a/curio/task.py
+++ b/curio/task.py
@@ -10,6 +10,13 @@ from collections import deque
 import linecache
 import traceback
 import os.path
+from functools import partial
+
+try:
+    import contextvars
+    _has_contextvars = True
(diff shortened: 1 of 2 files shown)


imrn:
Does it have anything to do with maintaining current_taak?

Fuyukai:
I personally think that they should be enabled and supported always, as it means there's no extra requirement for an app to use them other than use 3.7 or depend on contextvars for <3.6, instead of having to turn them on yourself (also can interact with libraries weirdly).

dabeaz:
Ok, I'm sold!

dabeaz:
I had to revert this pull request since it broke some stuff with async threads.  I'm going to revisit this, but the implementation will be different.

PR #280, opened 2018-10-08
add join for tasks in semaphore example

alekibango:
Without joining those tasks the program runs much
faster than expected by novice users.

This change makes  the example work as expected.

I also added more verbose text output,
showing state of the semaphore.


dabeaz:
Good catch!  Yes, it should definitely wait for them to finish.

PR #297, opened 2019-04-25
curio.run(coro, shutdown=True)

nikhilro:
Hey David (@dabeaz), we would sometimes get [this `RuntimeError`](https://github.com/dabeaz/curio/blob/master/curio/kernel.py#L121) when something deep in one of our tasks went awry. It was resolved by using `curio.Kernel(...).run(coro, ..., shutdown=True)`. `curio.run` should expose that control since it seems natural to be unaware of the Kernel and mostly only rely on the helper function. Thanks!

### DIFF
--- a/curio/kernel.py
+++ b/curio/kernel.py
@@ -816,7 +816,7 @@ class Kernel(object):
                     current = active = None
 
 def run(corofunc, *args, with_monitor=False, selector=None,
-        debug=None, activations=None, **kernel_extra):
+        debug=None, activations=None, shutdown=False, **kernel_extra):
     '''
     Run the curio kernel with an initial task and execute until all
     tasks terminate.  Returns the task's final result (if any). This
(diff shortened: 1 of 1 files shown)


dabeaz:
The global `run()` function executes code using the kernel as a context manager.  The context manager already executes `Kernel.run(shutdown=True)`.  Adding a `shutdown` option to `run()` would be redundant.  Perhaps I'm missing something.

nikhilro:
Ahh yes, you're right. I checked my local version and turns out I was using `v0.9` from pip with [this](https://github.com/dabeaz/curio/blob/0.9/curio/kernel.py#L127) as `__exit__`. Simply changing those two lines in `__exit__` with the now correct version properly shuts down. Apologies, for my laziness in just quickly glancing over the codebase. Will close this now. Thanks

PR #298, opened 2019-05-02
Travis: Run tests on Windows too

bauerj:
Hey,

I modified the Travis configuration slightly to also run the test suite on Windows. As you can see after merging this, some of the tests currently fail on Windows (they do on Travis and on my computer).

I'm not sure if the project officially supports Windows - the documentation doesn't mention anything about that as far as I can see. If it's not supported, just close this PR and maybe mention that somewhere. Otherwise I think it would be a good idea to have an overview of how the tests behave on Windows.

Let me know if you would rather have the Windows build as "allowed failure" so it doesn't cause the entire job to fail.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,13 +3,22 @@ matrix:
     - language: python
       python: 3.6
       os: linux
+    - os: windows
+      language: sh # Python is not yet officially supported
+  allow_failures:
+    - os: windows
+      language: sh
+  fast_finish: true
 
 sudo: false
 cache: pip
 
 before_install:
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
-  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew install python3; fi
+  - >
+    if [[ "$TRAVIS_OS_NAME" == "windows" ]]; then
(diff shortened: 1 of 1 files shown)


dabeaz:
Windows is not an officially supported platform to the extent that all of the development takes place on OSX/Linux.  I'm pretty sure there is a note about this being recommended in the README. 

That said, I'm not opposed to Curio working on Windows. None of the code is particularly specific to POSIX (with perhaps the exception of signal handling).   So, it wouldn't surprise me if most of it works already.

If you can make sure that the "allowed failure" option is set, I'd merge this.

bauerj:
I've changed the file accordingly. After being merged, it will look like this: https://travis-ci.org/bauerj/curio/builds/528704729

dabeaz:
Great!   I wonder if it might make sense to go through some of the tests and add skips for things that cause freezing/deadlock?     (Meanwhile, I'm just mentally thinking through some of the features that might cause problems besides signals).

bauerj:
Something like pytest-timeout would be nice, but it doesn't support resuming the test suite on Windows.

A quick hack to let all tests run is to use pytest-xdist and run e.g. 8 threads in parallel. Some threads still stall but the others continue the tests.

PR #302, opened 2019-08-15
Made kernel time handling simpler (and flatter)

GeeTransit:


GeeTransit:
By the way, if anyone is reading this, the new tkinter kernel example thing should probably be in its own pull request. I'll fork a new repository tomorrow to separate those two updates.

dabeaz:
Looks good. Would you be okay putting some kind of author attribution in the tkinter example?

GeeTransit:
I've added some attribution. (I'm famous now :D) Also, is it fine for the example program to be in the same pull request as the kernel time handling change or should they be separated?

dabeaz:
I'm ok with it all being a single PR.  Happy to have contributions!

PR #318, opened 2020-03-12
Mutable default arguments can cause issues

patrickBerry:
Default arguments that are mutable are bad because they are the same object and one modification can affect the next call and cause unintended behavior. If this was the intended behavior, then at least a comment should be added stating that such behavior was intended.

### DIFF
--- a/curio/thread.py
+++ b/curio/thread.py
@@ -67,10 +67,10 @@ _locals = threading.local()
 
 class AsyncThread(object):
 
-    def __init__(self, target=None, args=(), kwargs={}, daemon=False):
+    def __init__(self, target=None, args=None, kwargs=None, daemon=False):
         self.target = target
-        self.args = args
-        self.kwargs = kwargs
+        self.args = args if args is not None else tuple()
+        self.kwargs = kwargs if kwargs is not None else {}
         self.daemon = daemon
 
         # The following attributes are provided to make a thread mimic a Task


dabeaz:
Where are they mutated?

patrickBerry:
Not sure really. I ran the deepsource.io GitHub plugin on a project of mine that uses Curio (big fan btw) and found this as one of the reports. There were a number of other possible issues reported, but this issue seemed the easiest to report for a pull request.

dabeaz:
I've looked at the code and the only use of args and kwargs is as a pass-through to a func(*args, **kwargs) call.  There is no mutation.  As a general rule, I don't like complicating code to account for "what-if" scenarios that never happen.

curio/thread.py
```diff
@@ -67,10 +67,10 @@
 
 class AsyncThread(object):
 
-    def __init__(self, target=None, args=(), kwargs={}, daemon=False):
+    def __init__(self, target=None, args=None, kwargs=None, daemon=False):
```
jab: @patrickBerry, `()` is already immutable, so that half of this PR is invalid.

The `kwargs={}` half might actually be valid, in general, since that dict instance is essentially a module global that anyone who can get a reference to it can mutate for everyone else.
```python
>>> class Foo:
...     def __init__(self, kwargs={}):
...         self.kwargs = kwargs
... 
>>> f1 = Foo()
>>> f2 = Foo()
>>> f2.kwargs['global'] = True
>>> f1.kwargs
{'global': True}
```
But rather than changing this to `kwargs=None`, it seems cleaner to use `**kwargs`. Then each call that doesn't supply any kwargs results in `kwargs` getting bound to a new empty dict, resolving this concern, and the code on line 73 doesn't actually have to change.

PR #328, opened 2020-07-12
Fixed pytest plugin wrapping functions too eagerly

### LINKED ISSUE #327: Pytest plugin conflicts with AnyIO's
Curio's pytest plugin wraps any test coroutine function that contains the keyword `curio` anywhere. This causes problems for the AnyIO pytest plugin which, when testing against Curio, does cause `curio` to appear on this list. The fix is to do:
```python3
if pyfuncitem.get_closest_marker('curio'):
```
instead of
```python3
if 'curio' in pyfuncitem.keywords:
```
in `curio/pytest_plugin.py`.

This of course would be easier to test if the async tests in curio's test suite were coroutine functions and relied on curio's own pytest plugin to run them.

agronholm:
Using "keywords" here conflicts with the AnyIO pytest plugin.

Fixes #327.

### DIFF
--- a/curio/pytest_plugin.py
+++ b/curio/pytest_plugin.py
@@ -66,7 +66,7 @@ def pytest_pycollect_makeitem(collector, name, obj):
 def pytest_pyfunc_call(pyfuncitem):
     """Run curio marked test functions in a Curio kernel instead of a normal function call.
     """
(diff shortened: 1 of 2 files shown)


agronholm:
Thanks!

PR #354, opened 2022-07-07
Fix for windows OSError raised when select is called with empty lists since python 3.10.5

stharding:
python3.10 sets e.errno to 22 instead of 10022 so we need to check against both WSAEINVAL and EINVAL


pquentin:
For reference, this also applies to Python 3.9.13 and later as the fix was backported: https://docs.python.org/3.9/whatsnew/changelog.html#python-3-9-13-final

PR #376, opened 2025-10-14
Fixes #365: Don't pop an empty queue

### LINKED ISSUE #365: IndexError: pop from an empty deque from _kernel_wake
Got this exception from inside curio/sched.py

Not sure yet as to the cause.
```
Traceback (most recent call last):
  File "...", line 72, in <module>
    curio.run(main)
  File "...\venv\Lib\site-packages\curio\kernel.py", line 824, in run
    return kernel.run(corofunc, *args)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\curio\kernel.py", line 146, in run
    task = self._runner(coro)
           ^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\curio\kernel.py", line 768, in kernel_run
    traps[trap[0]](*trap[1:])
  File "...\venv\Lib\site-packages\curio\kernel.py", line 519, in trap_sched_wake
    tasks = sched._kernel_wake(n)
            ^^^^^^^^^^^^^^^^^^^^^
  File "...\venv\Lib\site-packages\curio\sched.py", line 88, in _kernel_wake
    task, = self._queue.popleft()
            ^^^^^^^^^^^^^^^^^^^^^
IndexError: pop from an empty deque
```

mwerezak:
Fix IndexError from being raised inside the kernel by turning SchedFIFO wake into a no-op if there are no tasks waiting.

See discussion on #365.

### DIFF
--- a/curio/sched.py
+++ b/curio/sched.py
@@ -84,7 +84,7 @@ class SchedFIFO(SchedBase):
 
     def _kernel_wake(self, ntasks=1):
         tasks = []
-        while ntasks > 0:
+        while self._queue and ntasks > 0:
             task, = self._queue.popleft()
             if task:
                 tasks.append(task)
--- a/CHANGES
+++ b/CHANGES
@@ -10,6 +10,8 @@ may have moved code there.  Many features of Curio were simply
 higher-level modules implemented on top of an existing core and can be
 added back to your code without much effort. -- Dave
 
+08/19/2024 Export Connection class to help with type hinting.
+
 04/11/2024 Removed undocumented ProcessPool and ThreadPool names.
 
 04/11/2024 Removed block_in_thread().  This functionality can be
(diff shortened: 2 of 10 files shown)


curio/sched.py
```diff
@@ -84,7 +84,7 @@ def remove():
 
     def _kernel_wake(self, ntasks=1):
         tasks = []
-        while ntasks > 0:
+        while ntasks > 0 and len(self._queue) > 0:
```
x42005e1f: I think that truth testing will also be sufficient:

```python
while self._queue and ntasks > 0:
    ...
```

This is the same as checking the length for non-zero (it cannot be less than zero anyway), but it takes fewer bytecode instructions. And this pattern is also present in the existing code base.

